PR #7, opened 2015-09-01
iterator interface

SimonDanisch:
nice for `...` and `join(::Colerant, " ")`

### DIFF
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -259,3 +259,7 @@ end
 
 zero{T}(::Type{Gray{T}}) = Gray{T}(zero(T))
  one{T}(::Type{Gray{T}}) = Gray{T}(one(T))
+
+Base.start(A::Colorant)                = 1
+Base.next(A::Colorant, state::Integer) = (A.(state), state+1)
+Base.done(A::Colorant, state::Integer) = length(A) < state


timholy:
I guess this gets back to the age-old question: is a 3-component color a single object or a vector-like container? I'm not necessarily opposed to having this, but it would be interesting to see examples of code that this simplifies.

SimonDanisch:
Ah totally forgot about this. Well I'm using `...` to extract the values there and there. E.g.
r, g, b = rgbcolor
I just used min/max/maximum/minimum for some color manipulation, which are more on the vector side.
And `join` as I already point out. I use it for IO. It yields more homogenous code to handle colors and vectors in the same way in that case.

timholy:
But what should it do for `BGR`? Should it give `b, g, r`? Or, like the constructor, `r,g,b`? One thing I don't like about this is that it's ambiguous whether you want the internal representation or the meaning. It would be a little bit like if we decided to get clever and make julia do this:

```jl
for (i,bit) in enumerate(5)
    println("The $(i)th bit is $bit")
end
```
Then we'd have to worry about whether this was a little or big endian machine. Not to mention all the bugs people would get who triggered `BitIterator` by accidentally writing `for i in length(A)`.

SimonDanisch:
> It would be a little bit like if we decided to get clever and make julia do this:

I would never do `5[1]`, but I do `rgb.r` in almost any of my color code.
Also, `lenght(5)` returns `1`, while `length(rgb)` returns `3`, implying a different iterator behavior.

I think I'm slowly starting to understand why people insist, that we should treat colors as scalars.
It's because in the frequency domain a color is a scalar value.
But isn't what we do in computer science, that we transform it into a color space, with a different basis? And then it's pretty clearly a vector? And we do this, because it's the easiest way for humans to handle and think about colors?
From the first time using computers with Photoshop, where you choose a color via 3-4 values or on a 2D map + slider, until doing Computer vision with k-means and color spaces, I've always treated colors mentally as vectors. Now with `GLSL` this is further enforced, since colors are exclusively treates as vectors.
It looks to me like trying to ride two horses at the same time, when we try to enforce the notion of color as a scalar, but in reality we have already transformed it to be in a vector space.

Call me less pedantic, but I would just iterate in the order it lays in memory. That's pretty much what we do for matrices, even though that column and row order are equally justified (so memory layout is taken as the decision criterion).

timholy:
> but I do rgb.r in almost any of my color code.

Better to write `red(rgb)`, because if `rgb` is an `RGB24`, it doesn't have any fields.

> Also, `lenght(5)` returns `1`, while `length(rgb)` returns `3`, implying a different iterator behavior.

Good point. If we decide against making them iterable, we should probably use a different name for that operation. If we decide in favor of making them iterable, it can stay.

> I think I'm slowly starting to understand why people insist, that we should treat colors as scalars.
It's because in the frequency domain a color is a scalar value.

But it's less about wavelength (I think it's correct to say that not all colors can be generated with a single wavelength) and more about how I think about pixels. I don't really want to think about an image as an array of arrays, I want to think about it as if it's an array of values.

> Call me less pedantic, but I would just iterate in the order it lays in memory. That's pretty much what we do for matrices, even though that column and row order are equally justified (so memory layout is taken as the decision criterion).

But then you will get bugs from code like this:
```
rgb = A[15,20]
r, g, b = rgb
B[15,20] = RGB(r,g,b)
```
if `A` is a BGR-array. You'd have to write separate versions of your function for each `AbstractRGB` type. Yuck.

timholy:
Regarding matrices: we don't have row-major matrices, and the main reason is because it would be such a pain-in-the-neck to implement well because our algorithms were never written with an eye towards different storage orders. In contrast, I tried to plan ahead with Colors, and make sure we were storage-order independent. In my opinion, making colors iterable is a big threat to that.

SimonDanisch:
I think it's a reasonable assumption, that when you iterate you must know what color type you have at hand. If you're writing code that can't make that assumptions, then you can still use `red`/`green`/etc. 
Should we really disallow orthogonal behaviour, just because you can't apply it everywhere?

> If we decide against making them iterable, we should probably use a different name for that operation.

That would introduce a few different code paths for me, so I'd cry a little ;)

timholy:
I suppose we could try it. You're right that one doesn't have to use the feature; the biggest risk is whether it will trap people by accident. If you're gung-ho to see how it works, go ahead and merge and let's see how it works out.

SimonDanisch:
This might be more pleasant:
```Julia
const Val1 = Val{1}()
const Val2 = Val{2}()
const Val3 = Val{3}()
const Val4 = Val{4}()
const Val5 = Val{5}()

Base.start(A::Colorant)                = Val1
Base.next(A::Colorant, state::Val{1}) = (comp1(A), Val2)
Base.next(A::Colorant, state::Val{2}) = (comp2(A), Val3)
Base.next(A::Colorant, state::Val{3}) = (comp3(A), Val4)
Base.next(A::Colorant, state::Val{4}) = (alpha(A), Val5)
Base.done{N}(A::Colorant, state::Val{N}) = length(A) < N
```
However, it seems to be around 16 times slower even with more inlining and `-O`.

timholy:
It's a nice idea, but as you noted it will be slower. The reason is that with this design,
```
while !done(A, state)
    item, state = next(A, state)
    ...
end
```
`state` is changing its type on each iteration. If this loop were unrolled, that would not be a concern.

SimonDanisch:
Yeah, I hoped it would just unrole the loop...Well, LLVM is no wizzard
after all ;)

2015-09-04 12:23 GMT+02:00 Tim Holy <notifications@github.com>:

timholy:
One could write a macro...

timholy:
I realized there is a way to do this: write it as `for e in elements(c) ... end` and have `elements` create a
```jl
immutable FieldOrder{C<:Colorant,N}
    c::C
    order::NTuple{N,Int}
end
```
and then access the elements as `getfield(fo.c, fo.order[state])`.

kimikage:
The iterator has been implemented in PR #260, but I will not close this PR to refer to past discussions. Let's discuss API design in #190.

Issue #10, opened 2015-09-15
Problems with creating 10bit version of YCbCr


vchuravy:
A few of the types are only defined as 

``` julia
  YIQ{T<:FloatingPoint} <: Color{T,3
  YCbCr{T<:FloatingPoint} <: Color{T,3}
```

Would it be appropriate to change these type bounds to `Fractional` instead?

The idea is coming from #9 to define:

``` julia
typealias YCbCr10 YCbCr{Ufixed10}
typealias Gray10 Gray{Ufixed10}
```

I can create a PR for doing this, but I don't know for which formats this would be appropriate and how to fix the tests.

``` julia
julia> Pkg.test("ColorTypes")
INFO: Testing ColorTypes
ERROR: LoadError: test failed: ColorTypes.YCbCr{T<:AbstractFloat} == ColorTypes.YCbCr{T<:Union{AbstractFloat,FixedPointNumbers.FixedPoint}}
 in expression: color_type(A) == C
 in default_handler at ./test.jl:30
 in do_test at ./test.jl:53
 [inlined code] from /home/wallnuss/.julia/v0.4/ColorTypes/test/runtests.jl:109
 in anonymous at ./no file:0
while loading /home/wallnuss/.julia/v0.4/ColorTypes/test/runtests.jl, in expression starting on line 106
```

timholy:
For anything where the "gamut" (the allowed space of values) fit in the proposed element type, it would definitely be good to broaden the bounds. I didn't take the time to figure out the proper gamut for all the color types; I did look into `XYZ`, and a few colors can generate `XYZ` values bigger than 1, so I left it `FloatingPoint`.

timholy:
In general ColorTypes hasn't been exporting `typealiases` thus far, since they are easy to recreate in packages. But we could change that.

vchuravy:
I think we don't need a lot of different typealiases, what we might need though is something along the way of `AbstractYCrCb` to allow compact storage formats.

vchuravy:
So I think adding Fractional makes only sense for `YIQ` and `YCbCr`  the other have different ranges for their components

vchuravy:
Other problems I am currently running into is that the `correct_gamut` and `cnvt` function in `Colors` are expecting 8bit values.

timholy:
Clearly that will have to fixed. Feel free to try whatever fix works for your more general case.

vchuravy:
[...]

PR #11, opened 2015-09-16
[RFC] widen type bound for YCbCr

vchuravy:
See #10 for Context.

These are the minimal changes to make YCbCr work with Fractional as a upper bound.

I think there is a case to introduce AbstractYCbCr to furthermore allow the introduction of compact types like RGB32 and also because I had to treat YCbCr like AbstractRGB in a few places.

The only point that I see that speaks against AbstractYCbCr is that it furthermore complicates the type hierarchy and it probably needs more consideration then these changes.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -245,7 +245,7 @@ immutable YIQ{T<:FloatingPoint} <: Color{T,3}
 end
 
 @doc "`YCbCr` is the Y'CbCr color encoding often used in digital photography or video" ->
-immutable YCbCr{T<:FloatingPoint} <: Color{T,3}
+immutable YCbCr{T<:Fractional} <: Color{T,3}
     y::T
     cb::T
     cr::T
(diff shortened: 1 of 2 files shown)


timholy:
I don't really know what I'm talking about, but don't Cb and Cr range from -1 to 1? https://en.wikipedia.org/wiki/File:YCbCr-CbCr_Scaled_Y50.png. If so, then you'll want `FixedPoint` rather than `Ufixed`.

vchuravy:
So I think there is Cb' / Cr' which go from 0-1 and alternatively Cb / Cr which go from -1 - 1. And technically the standard is speaking of DY' and DCb' + DCr', but I also don't know what I am talking about ;) 

I am trying to find the right standard definition right now... 

But Y' is definitely only from 0-1 so how would we handle a colortype that has two different value types?

vchuravy:
Also sub-black information encoding? Who came up with this...

vchuravy:
But you are right for Cb and Cr (8bit) 128.00b is defined as zero.

AndrewGibb:
We use sub-black in broadcast for encoding additional information in the physical layer on analogue networks like HD-SDI. I'd like to say it's a throwback to earlier times, but very few manufacturers of broadcast kit have dared remove HD-SDI connectors from their gear. My colleagues are working on an IP(as in Internet Protocol)-only broadcast system, but it is still very much a research project. So, sadly, sub-black info is around for a while. I suspect I'll either build a custom colour type for broadcast data, or more likely convert to floating point when I need to do any of the kind of processing that these libraries are intended for.

vchuravy:
@AndrewGibb That's what I need it for to process video material coming in from a Camera via SDI.

PR #14, opened 2015-09-20
Making the @doc macro work

mbeltagy:
After doing a Pkg.update() I ran into problem running PyPlot, see  https://github.com/stevengj/PyPlot.jl/issues/169

I pinned it down to ColorTypes.jl 

```julia
using ColorTypes
```
```
ERROR: @doc not defined
 in include at ./boot.jl:245
 in include_from_node1 at ./loading.jl:128
 in reload_path at loading.jl:152
 in _require at loading.jl:67
 in require at loading.jl:51
while loading /home/elbeltagy/.julia/ColorTypes/src/ColorTypes.jl, in expression starting on line 86
```

This changed fixed it.

### DIFF
--- a/src/ColorTypes.jl
+++ b/src/ColorTypes.jl
@@ -50,7 +50,7 @@ include("show.jl")
 end # module
 
 if VERSION < v"0.4.0-dev"
-    using Docile
+    using Docile, Compat
 end
 
 @doc """


timholy:
Thanks!

src/ColorTypes.jl
```diff
@@ -50,7 +50,7 @@ include("show.jl")
 end # module
 
 if VERSION < v"0.4.0-dev"
-    using Docile
+    using Docile, Compat
```
stevengj: Really, the `using Compat` should be unconditional (not inside an `if` statement).  It is innocuous in 0.4, gives compatibility with some 0.4 pre-releases, and will soon be needed for 0.5 compatibility anyway.

mbeltagy: I noticed this line change is totally unnecessary now. You can revert to the old code.
This happened after Compat had a version update a few hours ago. 
https://github.com/JuliaLang/METADATA.jl/commit/f38bcc2c6b1feab55b010bb7b7a322e9fb52cd6a

Issue #16, opened 2015-10-01
Finding the definition of RGBA


andrewcooke:
it's taken me a good 45 minutes to work out how RGBA is defined (a lot of that spent grepping other packages wondering what i was doing wrong because i didn't see it in ColorTypes).  maybe i shouldn't be doing that, or maybe i'm dumb, but i do worry that the code is "too clever".

i don't have a good solution for how you improve things, sorry.  but i felt i should flag this.  i've written similar code in my life and, in general, come to regret it when other people don't understand it.

SimonDanisch:
I run into the same thing and am not too happy about it, either.
We should write out at least the types and exports.

SimonDanisch:
I put it on my list at low priority. If you have some time feel free to prepare a PR.
Probably wait for @timholy's comment, since he wrote the code.
Best,
Simon

timholy:
I'd be fine with a PR that did all this manually.

andrewcooke:
yikes.  i am not going to do all that manually!  how about https://github.com/JuliaGraphics/ColorTypes.jl/pull/17 ?  it makes the names explicit, so that they can be found more easily.  one downside is that the explicit list needs to be maintained, so i added a tets.

timholy:
Looks great, thanks!

PR #17, opened 2015-10-01
Make types explicit in macro call

andrewcooke:
See https://github.com/JuliaGraphics/ColorTypes.jl/issues/16


timholy:
Now I understand better what you are after. Yes, this is a good addition. Thanks!

PR #46, opened 2016-08-15
Support "alpha swaps" in alphacolor & coloralpha

timholy:
Shh, if I commit this quickly, no one will notice that I had a longstanding bug (untested, naturally) in [these](https://github.com/JuliaGraphics/ColorTypes.jl/blob/17ad370f23dc0425d9894f4c842d3f41f62dc504/src/conversions.jl#L61) [lines](https://github.com/JuliaGraphics/ColorTypes.jl/blob/17ad370f23dc0425d9894f4c842d3f41f62dc504/src/conversions.jl#L64). (`color_type` should have just been `color`.) And I can pretend it's all about supporting a new feature, swapping in a new alpha value.


Evizero:
huh, this went right by me.

timholy:
Looks like Travis got stuck. I took it as an opportunity to turn on testing for 0.5 (nightly == 0.6-dev now).

timholy:
I'm gonna merge this, Travis just seems to be having trouble today (and 11/12 workers passed).

PR #47, opened 2016-08-15
Increase test coverage

Evizero:
🆙 ... for fun and profit! (minus the profit) 

- fix `show` issue for `AGray32` with custom `comp1` method
- add `alphacolor` for `Gray24` and `RGB24`
- make `HSB` a typealias for `HSV` instead of a function
- remove `convert(::Type{ARGB32},x::Real,alpha)=ARGB32(x,alpha)` (because 404: corresponding constructor not found)
- remove interestingly placed white-spaces (because I had to look at them for too long and couldn't resist anymore :-) )


src/conversions.jl
```diff
@@ -47,20 +47,20 @@ convert(::Type{RGB24},   x::Real) = RGB24(x)
 convert(::Type{ARGB32},  x::Real) = ARGB32(x)
 convert(::Type{Gray24},  x::Real) = Gray24(x)
 convert(::Type{AGray32}, x::Real) = AGray32(x)
-convert(::Type{ARGB32},  x::Real, alpha) = ARGB32(x, alpha)
```
Evizero: I don't understand how this should have worked, so I removed it (since I was unable to find a way to test it). Please correct me if I misunderstood (In that case I could use a few hints what to do here)

src/types.jl
```diff
@@ -144,7 +144,7 @@ immutable HSV{T<:AbstractFloat} <: Color{T,3}
 end
 
 "`HSB` (Hue-Saturation-Brightness) is an alias for `HSV`."
-HSB(h, s, b) = HSV(h, s, b)
+typealias HSB HSV
```
Evizero: Was there a reason to not do a typealias that I did not consider? This line was untested, so I took the easy way :-)

src/conversions.jl
Evizero: For some reason I have the sneaking suspicion that removing this method is causing the 0.5 error

Evizero:
I'm not sure why it would work on 0.4 but not 0.5. I will have to come back to this after I have upgraded to 0.5 to figure this out

src/types.jl
timholy: No, that seems reasonable.

timholy:
If you use fixed-width, those whitespaces made the braces line up. But you're not the first person to find them "interesting," so we can delete them.

test/runtests.jl
```diff
@@ -182,22 +200,36 @@ ac2 = convert(ARGB32, c)
...
     @test argb.r == red(ac)
     @test argb.g == green(ac)
     @test argb.b == blue(ac)
+    @test rgba.alpha == alpha(ac)
+    @test rgba.r == red(ac)
+    @test rgba.g == green(ac)
+    @test rgba.b == blue(ac)
 end
 
+@test convert(Float64, Gray(.3)) == .3
```
timholy: You can combine this and the next line with `===` (3 equals signs), which requires that the left and right sides be indistinguishable (including type).

timholy:
This looks great to me. I made one comment just to prove I read through it.

timholy:
With regards to the failure on 0.5 and higher, I guess that constructors don't fall back to `convert` so readily on 0.5? Perhaps especially for 2-argument versions? You might consider adding a specialized `Color` constructor [here](https://github.com/JuliaGraphics/ColorTypes.jl/blob/88a9cd44534c078471cb2bc260c974f1b14bae74/src/types.jl#L286-L289).

Evizero:
sounds easy enough, I'll boot back up an try that

Evizero:
green lights! although I quietly suspect that we just suppressed the symptoms of a deeper issue here

timholy:
Possibly. I'm unsure myself.

PR #49, opened 2016-09-19
Correct InexactError in tests due to FPN widening

timholy:
https://github.com/JeffBezanson/FixedPointNumbers.jl/pull/48 caught an error in the tests in this package. CC @Evizero. (That result must have seemed strange to you! Sorry I didn't catch it in my review.)

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -295,13 +295,13 @@ end
 # UInt32 comparison
 @test Gray24() == 0x00000000
 @test Gray24(.2) == 0x00333333
-@test Gray24(0xdd) == 0x00232323
-@test convert(UInt32, Gray24(0xdd)) === 0x00232323
+@test Gray24(0x23uf8) == 0x00232323
+@test convert(UInt32, Gray24(0x23uf8)) === 0x00232323
 @test AGray32() == 0xff000000
 @test AGray32(.2) == 0xff333333
 @test convert(AGray32, .2, 0.) == 0x00333333
-@test AGray32(0xdd) == 0xff232323
-@test convert(UInt32, AGray32(0xdd)) === 0xff232323
+@test AGray32(0x23uf8) == 0xff232323
+@test convert(UInt32, AGray32(0x23uf8)) === 0xff232323
 @test RGB24() == 0x00000000
 @test RGB24(0x00232323) == 0x00232323
 @test convert(UInt32, RGB24(0x00232323)) === 0x00232323
@@ -309,4 +309,3 @@ end
 @test ARGB32(0xff232323) == 0xff232323
 @test ARGB32(1,.2,.3) == 0xffff334c
 @test convert(UInt32, ARGB32(1,.2,.3)) === 0xffff334c
-


Evizero:
how did the tests pass in the first place?

timholy:
What happened before was the analog of the following:
```jl
julia> 0xdd*0xff
0x23
```
whereas now FixedPointNumbers does that conversion like this:
```jl
julia> 0xdd*0x00ff
0xdc23
```
which is bigger than `0xff` so it can't be reinterpreted as a `UFixed8`. In other words, the test was sensitive to (and indeed tested for) arithmetic overflow.

timholy:
Stated another way, `UFixed8(x)` takes `x` at face value, which means `x` must be between 0 and 1 to be convertible into a `UFixed8`. Internally, such numbers *are* represented as `UInt8`, so it's a common mistake (one I've made many times myself) to think that `0xdd` is equivalent to `0xdduf8`, but that would be inconsistent: it's `0xdd/0xff ≈ 0.867` that's equivalent to `0xdduf8`, and `0xdd == 221` is far out-of-bounds. If we made a special exception for `UInt8`, it would mean that `UFixed8(1) == 1` but `UFixed8(UInt8(1)) ≈ 0.004`. Talk about havoc :smile:.

The new FPN behavior should make it much easier for us all (myself included) to think about this clearly.

Evizero:
Got it. Thanks for explaining. So I take it `@assert Gray24(0x00232323) == 0x00232323` should now work? (similar to the ARGB tests below)?

Evizero:
Or is the idea to use `reinterpret` in such usecases in the future

timholy:
> So I take it @assert Gray24(0x00232323) == 0x00232323 should now work?

You might think so, but `Gray24` only has 8 bits for each color channel, and `0x00232323 ≈ 2.302755e6` is far too big to represent with 8 bits. So `Gray24(0x00232323)` will need to throw an error. You can do this, though:
```jl
julia> Gray(0x00232323)
ERROR: InexactError()
 in ColorTypes.Gray{T<:Union{AbstractFloat,Bool,FixedPointNumbers.FixedPoint}}(::UInt32) at /home/tim/.julia/v0.5/ColorTypes/src/types.jl:368

julia> Gray(Float64(0x00232323))
Gray{Float64}(2.302755e6)
```
For a type that *can* represent numbers beyond 0 and 1, it's actually quite nice that we allow it, because it makes things like computing the mean intensity across color channels easy (accumulate to a `RGB{Float64}` or `Gray{Float64}` and at the end divide by the total number of pixels). So it's not that we want/need to enforce that all colors are always between 0 and 1, we just want to make sure that types incapable of representing specific values throw errors rather than silently succeeding due to arithmetic overflow.

> Or is the idea to use reinterpret in such usecases in the future

Yes, `reinterpret` is perfect. Interestingly, it gives an error right now:
```jl
julia> reinterpret(Gray24, 0x00232323)
ERROR: reinterpret: target type not a leaf bitstype
 in reinterpret(::Type{ColorTypes.Gray24}, ::UInt32) at ./essentials.jl:86

julia> isbits(Gray24)
true

julia> isleaftype(Gray24)
true
```
We should probably define a specific method for this. Note, of course, that while `reinterpret(Gray24, 0x00232323)` should do the obvious thing, it certainly won't compare as equal to `0x00232323` (but it will to `UFixed8(0.137)`).

timholy:
BTW, `Gray(0x002323)` throws an error is because we default to `UFixed8` for integers. It would seem excessive for
```jl
julia> RGB(1,0,0)
RGB{U8}(1.0,0.0,0.0)
```
to return a `RGB{Float64}`. This choice is of course up for grabs.

My preference is to keep it this way, but [improve the quality](https://github.com/JuliaMath/FixedPointNumbers.jl/pull/49) of our [error messages](https://github.com/JuliaLang/julia/issues/18521) so they give users helpful hints to make it easy to understand what's happening and how to fix it.

Evizero:
I see. Looking at the type definitions again right now. I remember I was wondering why the inner constructors were overwritten. Retracing my steps I guess my reasoning behind these tests was that if the parameter is of type `UInt32` that the value would be intended to just be used as the member variable as is, similar to what the default constructor for `ARGB32` does.

Evizero:
I have been a fan of defaulting to `RGB{U8}` by the way. Thanks again for taking the time to explain this to me.

PR #51, opened 2016-09-24
Throw informative errors for illegal construction

timholy:
I'm making a big push to make Images more newbie-friendly, and color handling is an important part of that. While at least some of us really like the fact that all (gray and RGB) color components lie in the range of 0 to 1, it's a perennial source of confusion for newbies. In https://github.com/JuliaMath/FixedPointNumbers.jl/pull/53 I added informative errors for conversion to fixed point, but I think it can be even nicer if we throw customized error messages here.

For example, the `InexactError` in https://github.com/JuliaGraphics/Colors.jl/issues/252 is now replaced with this:
```jl
julia> RGBA(56, 34, 23)
ERROR: ArgumentError: (56,34,23,1) are integers in the range 0-255, but integer inputs are encoded with the UFixed8
  type, an 8-bit type representing 256 discrete values between 0 and 1.
  Consider dividing your input values by 255, for example: RGBA{UFixed8}(56/255,34/255,23/255,1/255)
 in throw_colorerror(::Type{FixedPointNumbers.UFixed{UInt8,8}}, ::Tuple{Int64,Int64,Int64,Int64}) at /home/tim/.julia/v0.5/ColorTypes/src/types.jl:615
 in throw_colorerror(::Type{FixedPointNumbers.UFixed{UInt8,8}}, ::Int64, ::Int64, ::Int64, ::Int64) at /home/tim/.julia/v0.5/ColorTypes/src/types.jl:585
 in checkval at /home/tim/.julia/v0.5/ColorTypes/src/types.jl:575 [inlined]
 in ColorTypes.RGBA{FixedPointNumbers.UFixed{UInt8,8}}(::Int64, ::Int64, ::Int64, ::Int64) at /home/tim/.julia/v0.5/ColorTypes/src/types.jl:453
 in ColorTypes.RGBA{T<:Union{AbstractFloat,FixedPointNumbers.FixedPoint}}(::Int64, ::Int64, ::Int64) at /home/tim/.julia/v0.5/ColorTypes/src/types.jl:478
```
which I think is rather friendlier.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,4 @@ julia:
 notifications:
   email: false
 after_success:
-- julia -e 'cd(Pkg.dir("ColorTypes")); Pkg.add("Coverage"); using Coverage; Codecov.submit(process_folder())'
+-  julia -e 'if VERSION >= v"0.6.0-dev.565" cd(Pkg.dir("ColorTypes")); Pkg.add("Coverage"); using Coverage; Codecov.submit(Codecov.process_folder()); end'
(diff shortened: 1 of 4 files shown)


timholy:
I should add that this has been carefully engineered to have minimal performance impact. See details in https://github.com/JuliaLang/julia/issues/18521.

PR #53, opened 2016-09-27
equal for transparent colors

### LINKED ISSUE #52: BigFloat breaks == on `GrayA` and `AGray`
supprisingly the following comparisons do not work.
```julia
julia> AGray{BigFloat}(0.5,0.25) == AGray{BigFloat}(0.5,0.25)
false
julia> GrayA{BigFloat}(0.5,0.25) == GrayA{BigFloat}(0.5,0.25)
false
```
In both cases the comparison should return true.

SimonDanisch:
fix #52


timholy:
Thanks!

PR #57, opened 2016-10-08
use do-block form of mktemp to clean up the temporary file

tkelman:


timholy:
Ah, thanks for the correction!

PR #59, opened 2016-10-12
Do something sensible for `ccolor(Gray, Int)`

timholy:
On `master` we currently have this:
```jl
julia> using ColorTypes

julia> a = Gray[0.0]
1-element Array{ColorTypes.Gray,1}:
 Gray{Float64}(0.0)

julia> eltype(a)
ColorTypes.Gray{T<:Union{AbstractFloat,Bool,FixedPointNumbers.FixedPoint}}
```

Note that this is a container with abstract eltype. That's not ideal, and we should probably change it so that `Gray[0.0]` returns an array of concrete eltype. However, since it's possible to construct such an array directly with `a = Array{Gray}(1)`, the following bug still needs fixing:
```jl
julia> a[1] = 1
ERROR: MethodError: no method matching ccolor(::Type{ColorTypes.Gray}, ::Type{Int64})
Closest candidates are:
  ccolor{Cdest<:ColorTypes.Colorant{T,N},Csrc<:ColorTypes.Colorant{T,N}}(::Type{Cdest<:ColorTypes.Colorant}, ::Type{Csrc<:ColorTypes.Colorant}) at /home/tim/.julia/v0.5/ColorTypes/src/traits.jl:228
  ccolor{Cdest<:ColorTypes.Color{T,1},T<:Union{AbstractFloat,Bool,FixedPointNumbers.FixedPoint{T<:Integer,f}}}(::Type{Cdest<:ColorTypes.Color{T,1}}, ::Type{T<:Union{AbstractFloat,Bool,FixedPointNumbers.FixedPoint}}) at /home/tim/.julia/v0.5/ColorTypes/src/traits.jl:229
 in setindex!(::Array{ColorTypes.Gray,1}, ::Int64, ::Int64) at ./array.jl:415
```

And in this PR, it's fixed. We can still support julia 0.4, so I did so, but the extra effort it took means that 0.4's days are numbered as far as new updates go...

### DIFF
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -33,7 +33,7 @@ gray(c::Gray)    = c.val
 gray(c::TransparentGray) = c.val
 gray(c::Gray24)  = UFixed8(c.color & 0x000000ff, 0)
 gray(c::AGray32) = UFixed8(c.color & 0x000000ff, 0)
-gray(x::Union{Fractional,Bool}) = x
+gray(x::Number)  = x
 
 # Extract the first, second, and third arguments as you'd
 # pass them to the constructor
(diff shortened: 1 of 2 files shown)


timholy:
I'm contemplating turning off OSX from travis. The latency on OSX builds is really long these days. https://www.traviscistatus.com/ indicates that their load is pegged to the maximum except during the night.

Any objections?

tlnagy:
I'm not sure what is up with the OS X builds, but recently they are taking forever to initialize. @tkelman do you know any solution to this problem?

timholy:
I think it's just that their OSX build farm is maxed out---on the OSX plots, see the flat line on the number of concurrent jobs, whereas the linux servers clearly have more capacity because they never flatline at the top. So the initialization delay is just the latency for them to finish the jobs ahead of ours in the queue.

Assuming that's true, I think the answer is "send money to Travis." (I'm not trying to say that's an unreasonable request; it takes money to install more servers.)

tlnagy:
If all of their servers are timing out, it's not doing us much good to keep testing on OS X. I say turn it off.

tkelman:
They had another issue today. It's pretty frequent, and paying them on an open source project doesn't get you in a different queue, just more concurrent jobs.

I'd prefer keeping the platform coverage but ignore osx builds on days they have problems.

Another solution would be to complain to apple until they change their policy about virtualizing the os on non mac hardware for testing or development purposes, but I doubt they'd listen.

Issue #63, opened 2016-11-08
reinterpret problem


bjarthur:
`reinterpret` works for vector but not scalar `Gray{Float64}`.  it claims it's not a bitstype for the latter.  to debug, i made a fake `type Bar`, meant to mimic the definition of `Gray`, for which `reinterpret` works for neither scalar or vector.  could someone please give me a hint as how to fix this?  i do not see any applicable `reinterpret` methods in all of `ColorTypes.jl`.  thanks.

```
julia> using ColorTypes

julia> reinterpret(Gray{Float64},rand(Float64,3),(3,))
3-element Array{ColorTypes.Gray{Float64},1}:
 Gray{Float64}(0.807567) 
 Gray{Float64}(0.528818) 
 Gray{Float64}(0.0608983)

julia> reinterpret(Gray{Float64},rand(Float64))
ERROR: reinterpret: target type not a leaf bitstype
 in reinterpret(::Type{ColorTypes.Gray{Float64}}, ::Float64) at ./essentials.jl:86

julia> type Bar{T<:AbstractFloat}
              x::T
              end

julia> reinterpret(Bar{Float64},rand(Float64,3),(3,))
ERROR: ArgumentError: cannot reinterpret Array{Float64} to ::Type{Array{Bar{Float64}}}, type Bar{Float64} is not a bitstype
 in reinterpret(::Type{Bar{Float64}}, ::Array{Float64,1}, ::Tuple{Int64}) at ./array.jl:87

julia> reinterpret(Bar{Float64},rand(Float64))
ERROR: reinterpret: target type not a leaf bitstype
 in reinterpret(::Type{Bar{Float64}}, ::Float64) at ./essentials.jl:86
```

timholy:
`Bar` has to be an immutable.

The error just comes from the default `reinterpret` fallback. There's nothing wrong with defining
```jl
reinterpret{T}(::Type{Gray{T}}, x::T) = Gray(x)
```

bjarthur:
thanks for the help @timholy .

i'm inclined to think however that this is a bug in `Base`.  the `box` function that `reinterpret` calls for scalars differs in it's behavior to `Gray{Float64}` than the `jl_reshape_array` function that `reinterpret` calls for vectors.  see the simple hack below of passing in a vector of length one as a work around, and the different method that is called as a result.

[...]

PR #64, opened 2016-11-12
refactor rand to handle arrays of UFixed{UInt16,14} and friends

bjarthur:
it was hard coded for UFixed8/16 before, now is more general.

it's also a lot simpler, except the hack around https://github.com/JuliaLang/julia/issues/10140, which i'm not sure will add a lot of overhead or not.

### DIFF
--- a/src/operations.jl
+++ b/src/operations.jl
@@ -4,26 +4,64 @@ hash(c::TransparentGray, hx::UInt) = hash(alpha(c), hash(gray(c), hx))
 hash(c::AbstractRGB, hx::UInt) = hash(blue(c), hash(green(c), hash(red(c), hx)))
 hash(c::TransparentRGB, hx::UInt) = hash(alpha(c), hash(blue(c), hash(green(c), hash(red(c), hx))))
 
-# rand
-typealias RandTypes{T,C} Union{AbstractRGB{T},
-                               TransparentRGB{C,T},
-                               AbstractGray{T},
-                               TransparentGray{C,T}}
-typealias RandTypesFloat{C,T<:AbstractFloat} RandTypes{T,C}
+# gamut{min,max}
+gamutmax{T<:HSV}(::Type{T}) = (360,1,1)
+gamutmin{T<:HSV}(::Type{T}) = (0,0,0)
+gamutmax{T<:HSL}(::Type{T}) = (360,1,1)
+gamutmin{T<:HSL}(::Type{T}) = (0,0,0)
+gamutmax{T<:Lab}(::Type{T}) = (100,1,1)
+gamutmin{T<:Lab}(::Type{T}) = (0,0,0)
+gamutmax{T<:LCHab}(::Type{T}) = (100,1,360)
+gamutmin{T<:LCHab}(::Type{T}) = (0,0,0)
+gamutmax{T<:YIQ}(::Type{T}) = (1,0.5226,0.5226)
+gamutmin{T<:YIQ}(::Type{T}) = (0,-0.5957,-0.5957)
 
-rand{G<:AbstractGray}(::Type{G}) = G(rand())
-rand{G<:TransparentGray}(::Type{G}) = G(rand(), rand())
-rand{C<:AbstractRGB}(::Type{C}) = C(rand(), rand(), rand())
-rand{C<:TransparentRGB}(::Type{C}) = C(rand(), rand(), rand(), rand())
-function rand{C<:RandTypesFloat}(::Type{C}, sz::Dims)
-    reinterpret(C, rand(eltype(C), (sizeof(C)÷sizeof(eltype(C)), sz...)), sz)
+gamutmax{T<:AbstractGray}(::Type{T}) = (1,)
+gamutmax{T<:TransparentGray}(::Type{T}) = (1,1)
+gamutmax{T<:AbstractRGB}(::Type{T}) = (1,1,1)
+gamutmax{T<:TransparentRGB}(::Type{T}) = (1,1,1,1)
+gamutmin{T<:AbstractGray}(::Type{T}) = (0,)
+gamutmin{T<:TransparentGray}(::Type{T}) = (0,0)
+gamutmin{T<:AbstractRGB}(::Type{T}) = (0,0,0)
+gamutmin{T<:TransparentRGB}(::Type{T}) = (0,0,0,0)
+
+# rand
+for t in [Float16,Float32,Float64,N0f8,N0f16,N0f32]
+    @eval _rand{T<:Union{AbstractRGB{$t},AbstractGray{$t}}}(::Type{T}) =
+          mapc(x->rand(eltype(T)), base_colorant_type(T)())
 end
-function rand{C<:RandTypes{N0f8}}(::Type{C}, sz::Dims)
-    reinterpret(C, rand(UInt8, (sizeof(C), sz...)), sz)
+
+function _rand{T<:Colorant}(::Type{T})
+    Gmax = gamutmax(T)
+    Gmin = gamutmin(T)
+    Mi = eltype(T) <: FixedPoint ? 1.0/typemax(eltype(T)) : 1.0
+    A = rand(eltype(T), length(T))
+    for j in eachindex(Gmax)
+        A[j] = A[j] * (Mi * (Gmax[j]-Gmin[j])) + Gmin[j]
+    end
+    T(A...)
 end
-function rand{C<:RandTypes{N0f16}}(::Type{C}, sz::Dims)
-    reinterpret(C, rand(UInt16, (sizeof(C)>>1, sz...)), sz)
+
+function _rand{T<:Colorant}(::Type{T}, sz::Dims)
+    Gmax = gamutmax(T)
+    Gmin = gamutmin(T)
+    Mi = eltype(T) <: FixedPoint ? 1.0/typemax(eltype(T)) : 1.0
+    A = rand(eltype(T), (div(sizeof(T), sizeof(eltype(T))), sz...))
+    for j in eachindex(Gmax)
+        s = Mi * (Gmax[j]-Gmin[j])
+        for i = j:length(Gmax):length(A)
+            A[i] = A[i] * s + Gmin[j]
+        end
+    end
+    reinterpret(T, A, sz)
 end
+
+rand{T<:Colorant}(::Type{T}, sz::Dims...) = _rand(ccolor(T, base_colorant_type(T){Float64}), sz...)
+
+rand(::Type{Gray24}) = Gray24(rand(N0f8))
+rand(::Type{Gray24}, sz::Dims) = Gray24.(rand(N0f8,sz))
+rand(::Type{AGray32}) = AGray32(rand(N0f8),rand(N0f8))
+rand(::Type{AGray32}, sz::Dims) = AGray32.(rand(N0f8,sz),rand(N0f8,sz))
 
 # broadcast
 # Without this, Gray.(a) returns an Array{Gray}, which does not have concrete eltype
--- a/src/ColorTypes.jl
+++ b/src/ColorTypes.jl
@@ -42,7 +42,7 @@ export base_color_type, base_colorant_type, ccolor, color, color_type
 export alphacolor, coloralpha
 export alpha, red, green, blue, gray   # accessor functions that generalize to RGB24, etc.
 export comp1, comp2, comp3
-export mapc
+export mapc, gamutmax, gamutmin
 
 if VERSION < v"0.5.0-dev+1946"
     const supertype = super
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -413,17 +413,45 @@ Set([DIN99d, DIN99o, DIN99, HSI, HSL, HSV, LCHab, LCHuv,
      LMS, Lab, Luv, XYZ, YCbCr, YIQ, xyY, BGR, RGB, Gray])
 
 ## operations
-for T in (Gray{N0f8}, AGray{Float32}, GrayA{Float64},
-          RGB{N0f8}, ARGB{N0f16}, RGBA{Float32},
-          BGR{Float16}, RGB1{N0f8}, RGB4{Float64}, ABGR{N0f8})
+for T in (Gray{N0f8}, Gray{N2f6}, Gray{N0f16}, Gray{N2f14}, Gray{N0f32}, Gray{N2f30},
+          Gray{Float16}, Gray{Float32}, Gray{Float64},
+          RGB{N0f8}, RGB{N2f6}, RGB{N0f16}, RGB{N2f14}, RGB{N0f32}, RGB{N2f30},
+          RGB{Float16}, RGB{Float32}, RGB{Float64},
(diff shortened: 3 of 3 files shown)


bjarthur:
see discussion [here](https://github.com/JuliaGraphics/ColorTypes.jl/issues/63)

bjarthur:
gheesh, okay, this latest revision is now no longer 100x slower than master!  i really shouldn't code when i'm sick :(

interesting that `reinterpret(T...` is faster than `T(...`:

```
julia> randR{T<:Colorant}(::Type{T}) = reinterpret(T, rand(eltype(T),div(sizeof(T), sizeof(eltype(T)))))[1]

randR (generic function with 1 method)

julia> randT{T<:Colorant}(::Type{T}) = T( rand(eltype(T),length(T))...)
randT (generic function with 1 method)

julia> @benchmark randR(Gray{UFixed{UInt16,16}})
BenchmarkTools.Trial: 
  memory estimate:  432.00 bytes
  allocs estimate:  8
  --------------
  minimum time:     300.865 ns (0.00% GC)
  median time:      314.685 ns (0.00% GC)
  mean time:        478.179 ns (27.14% GC)
  maximum time:     34.473 μs (97.95% GC)
  --------------
  samples:          10000
  evals/sample:     252
  time tolerance:   5.00%
  memory tolerance: 1.00%

julia> @benchmark randT(Gray{UFixed{UInt16,16}})
BenchmarkTools.Trial: 
  memory estimate:  368.00 bytes
  allocs estimate:  7
  --------------
  minimum time:     355.607 ns (0.00% GC)
  median time:      372.107 ns (0.00% GC)
  mean time:        526.678 ns (21.98% GC)
  maximum time:     50.473 μs (98.45% GC)
  --------------
  samples:          10000
  evals/sample:     214
  time tolerance:   5.00%
  memory tolerance: 1.00%
```

lastly, the tests won't pass until https://github.com/JuliaMath/FixedPointNumbers.jl/pull/61 is merged and tagged.

bjarthur:
test are now passing on all but the nightly since https://github.com/JuliaMath/FixedPointNumbers.jl/pull/61 has been tagged.

could someone please merge this?  i don't have permissions.

https://github.com/JuliaGraphics/ColorVectorSpace.jl/pull/62 is waiting for this to be tagged.

src/operations.jl
```diff
@@ -5,25 +5,9 @@ hash(c::AbstractRGB, hx::UInt) = hash(blue(c), hash(green(c), hash(red(c), hx)))
...
-end
-function rand{C<:RandTypes{N0f8}}(::Type{C}, sz::Dims)
-    reinterpret(C, rand(UInt8, (sizeof(C), sz...)), sz)
-end
-function rand{C<:RandTypes{N0f16}}(::Type{C}, sz::Dims)
-    reinterpret(C, rand(UInt16, (sizeof(C)>>1, sz...)), sz)
-end
+rand{T<:Colorant}(::Type{T}) = reinterpret(T, rand(eltype(T), div(sizeof(T), sizeof(eltype(T)))))[1]
+rand{T<:Colorant}(::Type{T}, sz::Dims) =
+      reinterpret(T, rand(eltype(T), (div(sizeof(T), sizeof(eltype(T))), sz...)), sz)
```
timholy: Here and above: `rand(RGB)` works currently but won't when this is merged. Maybe consider
```julia
rand{T<:Colorant}(::Type{T}) = _rand(ccolor(T, base_colorant_type(T){Float64}))
```
and then having `_rand` be the "real" implementation.

src/operations.jl
```diff
@@ -5,25 +5,9 @@ hash(c::AbstractRGB, hx::UInt) = hash(blue(c), hash(green(c), hash(red(c), hx)))
...
-function rand{C<:RandTypesFloat}(::Type{C}, sz::Dims)
-    reinterpret(C, rand(eltype(C), (sizeof(C)÷sizeof(eltype(C)), sz...)), sz)
-end
-function rand{C<:RandTypes{N0f8}}(::Type{C}, sz::Dims)
-    reinterpret(C, rand(UInt8, (sizeof(C), sz...)), sz)
-end
-function rand{C<:RandTypes{N0f16}}(::Type{C}, sz::Dims)
-    reinterpret(C, rand(UInt16, (sizeof(C)>>1, sz...)), sz)
-end
+rand{T<:Colorant}(::Type{T}) = reinterpret(T, rand(eltype(T), div(sizeof(T), sizeof(eltype(T)))))[1]
```
timholy: Hmm, the performance implications are somewhat concerning. Maybe make the RHS
```julia
mapc(x->rand(eltype(T)), base_colorant_type(T)())
```
If we let `myrand` be yours and `myrand3` be mine, I get
```julia
julia> @benchmark myrand(HSV{Float32})
BenchmarkTools.Trial: 
  samples:          10000
  evals/sample:     759
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  352.00 bytes
  allocs estimate:  6
  minimum time:     155.00 ns (0.00% GC)
  median time:      164.00 ns (0.00% GC)
  mean time:        213.86 ns (21.27% GC)
  maximum time:     5.59 μs (93.40% GC)

julia> @benchmark myrand3(HSV{Float32})
BenchmarkTools.Trial: 
  samples:          10000
  evals/sample:     997
  time tolerance:   5.00%
  memory tolerance: 1.00%
  memory estimate:  0.00 bytes
  allocs estimate:  0
  minimum time:     21.00 ns (0.00% GC)
  median time:      22.00 ns (0.00% GC)
  mean time:        22.39 ns (0.00% GC)
  maximum time:     70.00 ns (0.00% GC)
```

src/operations.jl
bjarthur: `rand(RGB)` works for me locally, and it's in the tests, which pass.  how am i confused?

```
julia> rand(RGB{Float64},3)
3-element Array{ColorTypes.RGB{Float64},1}:
 RGB{Float64}(0.735732,0.836881,0.282634)
 RGB{Float64}(0.987138,0.619426,0.803145)
 RGB{Float64}(0.65319,0.329882,0.335785) 

julia> rand(RGB{Float64})
RGB{Float64}(0.5079139816107203,0.9483896058733416,0.40892847391611986)
```

timholy:
```julia
julia> using ColorTypes

julia> bja_rand{T<:Colorant}(::Type{T}) = reinterpret(T, rand(eltype(T), div(sizeof(T), sizeof(eltype(T)))))[1]
bja_rand (generic function with 1 method)

julia> bja_rand(RGB)
ERROR: argument is an abstract type; size is indeterminate
 in bja_rand(::Type{ColorTypes.RGB}) at ./REPL[2]:1

julia> rand(RGB)
RGB{Float64}(0.30389835762878525,0.1187732371310497,0.9873748322983826)
```
In the version I suggested, that `ccolor` call basically provides a "default fallback" of `Float64` (since that what `rand` actually produces).

bjarthur:
`mapc` performance optimization added for the scalar case.  is there a way to speed up arrays?

`RGB` without an eltype specified now works.

`HSV` explicitly not supported.

anything else?  tests are passing on julia 0.5.

src/operations.jl
```diff
@@ -5,25 +5,11 @@ hash(c::AbstractRGB, hx::UInt) = hash(blue(c), hash(green(c), hash(red(c), hx)))
...
-    reinterpret(C, rand(UInt8, (sizeof(C), sz...)), sz)
-end
-function rand{C<:RandTypes{N0f16}}(::Type{C}, sz::Dims)
-    reinterpret(C, rand(UInt16, (sizeof(C)>>1, sz...)), sz)
-end
+_rand{T<:Colorant}(::Type{T}) = mapc(x->rand(eltype(T)), base_colorant_type(T)())
+_rand{T<:Colorant}(::Type{T}, sz::Dims) =
+      reinterpret(T, rand(eltype(T), (div(sizeof(T), sizeof(eltype(T))), sz...)), sz)
+rand{T<:Colorant}(::Type{T}, sz::Dims...) = _rand(ccolor(T, base_colorant_type(T){Float64}), sz...)
+rand{T<:HSV}(::Type{T}, sz::Dims...) = error("rand(::Type{HSV}, sz...) not supported")
```
timholy: Sorry, I was just using HSV as an example. You'll have problems with lots of others too. For example:
```julia
julia> convert(DIN99d, rand(RGB))
DIN99d{Float64}(17.531226651912203,-1.7876008277942326,-27.780870734776467)
```
that single example goes way beyond the bounds of `[0,1]`. The fundamental question is, can `rand` sample the "gamut" of the colorspace? Even `LMS` and `XYZ` extend higher than 1.

timholy:
I don't know of a way to speed up arrays. Indeed `rand` has optimizations if you're asking for a bunch of random numbers at once, so given that you have to allocate an array anyway I suspect it's not possible to make better.

bjarthur:
in addition to working for e.g. `N2f14`, `rand` now scales appropriately for e.g. `HSV`.  moreover it special cases `AbstractRGB` and `AbstractGray` for performance.  and it works for `Gray24` and `AGray32`.

could someone please review?  thx.

src/operations.jl
```diff
@@ -4,25 +4,75 @@ hash(c::TransparentGray, hx::UInt) = hash(alpha(c), hash(gray(c), hx))
 hash(c::AbstractRGB, hx::UInt) = hash(blue(c), hash(green(c), hash(red(c), hx)))
 hash(c::TransparentRGB, hx::UInt) = hash(alpha(c), hash(blue(c), hash(green(c), hash(red(c), hx))))
 
+# range
+range{T<:HSV}(::Type{T}) = (360,1,1)
```
timholy: This seems like a rather different meaning of `range`. I think I'd prefer a separate name...`gamutmax`? Sort of like `typemax`?

src/operations.jl
```diff
@@ -4,25 +4,75 @@ hash(c::TransparentGray, hx::UInt) = hash(alpha(c), hash(gray(c), hx))
...
+range{T<:HSL}(::Type{T}) = (360,1,1)
+range{T<:Lab}(::Type{T}) = (100,1,1)
+range{T<:LCHab}(::Type{T}) = (100,1,360)
+typealias Colorant1{T} Colorant{T,1}
+typealias Colorant2{T} Colorant{T,2}
+typealias Colorant3{T} Colorant{T,3}
+typealias Colorant4{T} Colorant{T,4}
+range{T<:Colorant1}(::Type{T}) = (1,)
```
timholy: Misleading for color types like the `DIN` family. And `YIQ` can have negative components as well as positive, so the implicit `gamutmin` of 0 doesn't always hold.

src/operations.jl
```diff
@@ -4,25 +4,75 @@ hash(c::TransparentGray, hx::UInt) = hash(alpha(c), hash(gray(c), hx))
...
-function rand{C<:RandTypesFloat}(::Type{C}, sz::Dims)
-    reinterpret(C, rand(eltype(C), (sizeof(C)÷sizeof(eltype(C)), sz...)), sz)
+for t in [Float16,Float32,Float64,N0f8,N0f16,N0f32]
+    @eval _rand{T<:Union{AbstractRGB{$t},AbstractGray{$t}}}(::Type{T}) =
+          mapc(x->rand(eltype(T)), base_colorant_type(T)())
+end
+
+function _rand{T<:Colorant}(::Type{T})
+    R = range(T)
+    Mi = eltype(T) <: FixedPoint ? 1.0/typemax(eltype(T)) : 1.0
```
timholy: sweet. I hadn't thought about generating larger values for `N2f14`.

src/operations.jl
```diff
@@ -4,25 +4,75 @@ hash(c::TransparentGray, hx::UInt) = hash(alpha(c), hash(gray(c), hx))
...
+    A = rand(eltype(T), length(T))
```
timholy: Wouldn't `mapc((x,r)->r*Mi*rand(eltype(T)), base_colorant_type(T)(), R)` be considerably faster?

src/operations.jl
```diff
@@ -4,25 +4,75 @@ hash(c::TransparentGray, hx::UInt) = hash(alpha(c), hash(gray(c), hx))
...
+    for j in eachindex(R)
+        s = R[j] * Mi
+        for i = j:length(R):length(A)
+            A[i] *= s
+        end
+    end
+    reinterpret(T, A, sz)
+end
+
+rand{T<:Colorant}(::Type{T}, sz::Dims...) = _rand(ccolor(T, base_colorant_type(T){Float64}), sz...)
```
timholy: This splats `sz` but `_rand` above seems to want a tuple input.

src/operations.jl
```diff
@@ -4,25 +4,75 @@ hash(c::TransparentGray, hx::UInt) = hash(alpha(c), hash(gray(c), hx))
...
+
+function rand(::Type{Gray24})
+    x8 = rand(UInt8,1)
```
timholy: How about just `Gray24(rand(N0f8))`?

src/operations.jl
```diff
@@ -4,25 +4,75 @@ hash(c::TransparentGray, hx::UInt) = hash(alpha(c), hash(gray(c), hx))
...
+    x32 = repeat(x8, inner=(4,))
+    reinterpret(Gray24, x32)[1]
+end
+
+function rand(::Type{Gray24}, sz::Dims)
+    x8 = rand(UInt8,sz)
+    x32 = repeat(x8, inner=(4,ones(Int,length(sz)-1)...))
```
timholy: similarly, `Gray24.(x8)` where `x8 = rand(N0f8, sz)`. Same pattern below (I won't comment specifically).

src/operations.jl
bjarthur: can you provide an english language description of `DIN`?

src/operations.jl
bjarthur: this is the best i can come up with using `mapc`, and it's 2x slower:

```
function rand2{T<:Colorant}(::Type{T})
    Gmax = [gamutmax(T)...]
    Gmin = [gamutmin(T)...]
    Mi = eltype(T) <: FixedPoint ? 1.0/typemax(eltype(T)) : 1.0
    mapc((s,o)->rand(eltype(T))*s-o, T(Mi.*(Gmax-Gmin)...), T(Gmin...))
end
```

src/operations.jl
timholy: Nope, I don't really know what it's about. My main strategy would simply be to recommend *against* providing a generic fallback. Only have the method work for color types you want to support. Previously we only supported gray and RGB because those were the ones I cared about. This is much nicer, of course, but no need to kill yourself over types that people may not ever use `rand` with. After all, if they want random `DIN` colors, they can always say `convert(DIN99d, rand(RGB{Float32}))` or whatever.

src/operations.jl
timholy: Go with the fastest you have, then; if anyone cares they can optimize later. The general principle is: do everything with `()` (tuples) and not `[]` (arrays).

bjarthur:
tests are passing.  will merge in a couple days unless anyone has any other comments.

timholy:
Thanks, @bjarthur!

PR #68, opened 2017-02-01
WIP: 0.6 Compatibility

Keno:
Depends on https://github.com/JuliaLang/Compat.jl/pull/304
Fixes #65. Untested on 0.5 so far.

PR #69, opened 2017-02-07
Fix some breakages in #68

timholy:
Won't pass until https://github.com/JuliaLang/Compat.jl/pull/320 merges and tags. ~~Doesn't fix https://github.com/JuliaGraphics/ColorTypes.jl/pull/68#discussion_r99741490. Most other issues with #68 seem fixed, but see a specific comment below.~~

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -414,8 +414,16 @@ AGray32(g::AbstractGray, alpha = 1) = AGray32(gray(g), alpha)
 # Note: with the exceptions of `alphacolor` and `coloralpha`, all
 # traits in the rest of this file are intended just for internal use
 
-const color3types = filter(x->(!x.abstract && length(fieldnames(x))>1), union(subtypes(Color), subtypes(AbstractRGB)))
-const parametric3 = filter(x->!isempty(x.parameters), color3types)
+const color3types = map(s->getfield(ColorTypes,s),
+  filter(names(ColorTypes, true)) do s
+    isdefined(ColorTypes, s) || return false
+    t = getfield(ColorTypes, s)
+    isa(t, Type) && t <: Colorant && !isabstract(t) && length(fieldnames(t))>1
+  end)
+# The above should have filtered out every non-DataType that's not also a
+# wrapped UnionAll-wrapped DataType. By avoiding the explicit UnionAll check
+# here, we remain compatible with pre-0.6 julia.
+const parametric3 = filter(x->!isa(x, DataType) || !isempty(x.parameters), color3types)
 
 # Provide the field names in the order expected by the constructor
 colorfields{C<:Color}(::Type{C}) = (fieldnames(C)...)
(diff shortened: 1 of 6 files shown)


test/runtests.jl
```diff
@@ -79,8 +86,8 @@ a[1] = Gray(0)
 a[1] = 1
 @test a[1] === Gray(1)
 
-@test @inferred(ccolor(RGB,  RGB))  == RGB
-@test @inferred(ccolor(Gray, Gray)) == Gray
+# @test @inferred(ccolor(RGB,  RGB))  == RGB
+# @test @inferred(ccolor(Gray, Gray)) == Gray
```
timholy: Don't want to comment out these lines, but now we have this on julia-0.5:
```julia
julia> RGB
ColorTypes.RGB{T<:Union{AbstractFloat,FixedPointNumbers.FixedPoint}}

julia> ccolor(RGB, RGB)
ColorTypes.RGB{Union{AbstractFloat,FixedPointNumbers.FixedPoint}}

julia> ccolor(RGB, RGB) == RGB
false
```
Works fine on 0.6. Any idea what might have changed this?

Keno: Yes, I removed a generated function, because those are bad (TM). Then I fixed up inference to take care of this on 0.6. You can put back the generated function if you want (0.5 only).

Keno: Actually, I see you already did that. Maybe look for calls to parameter_upper_bound then. Those get the bounds not the typevar, so you may have to manually re-introduce that.

timholy: Thanks, that fixed it.

timholy:
I suspect we're good to go once https://github.com/JuliaLang/Compat.jl/pull/320 merges and tags. We'll need one more trip through CI.

test/runtests.jl
```diff
@@ -12,7 +12,18 @@ else
 end
 
 if VERSION >= v"0.5.0"
-    @test isempty(detect_ambiguities(ColorTypes, Base, Core))
+    if isdefined(Core, :UnionAll)
```
tkelman: not the right condition for the introduction of the allow_bottom kwarg which was introduced much later

tkelman:
Compat needs to be in REQUIRE if used

test/runtests.jl
timholy: ATM I think this branch is really about skipping the ambiguity test, which does coincide with `UnionAll`. When we uncomment that line, then I guess we have to be more careful.

src/types.jl
```diff
@@ -551,7 +559,7 @@ for (C, acol, cola) in [(DIN99d, :ADIN99d, :DIN99dA),
     cfn = Expr(:tuple, colorfields(C)...)
     elty = eltype_default(C)
     ub   = eltype_ub(C)
-    Csym = C.name.name
+    Csym = isdefined(Core, :UnionAll) ? Base.datatype_name(C) : C.name.name
```
tkelman: datatype_name on a UnionAll won't work properly in all versions where UnionAll was defined

timholy: Since ColorTypes was already completely hosed on 0.6 before this, and it hosed tests for ~50 packages, I don't think we need to wait to resolve this before merging https://github.com/JuliaLang/METADATA.jl/pull/7899.

tkelman: sure but it will widen the range of Julia versions on which this package works correctly, which is valuable for bisecting various issues

PR #72, opened 2017-02-11
Move one/zero from ColorVectorSpace

yuyichao:
Not sure how useful `one(Gray)` or `one(Gray{T} where T <: Fractional)` actually is but at least this solves the ambiguity error in ColorVectorSpace test.


timholy:
LGTM. We should add `oneunit` to Compat and start using that instead, but that's a separate PR.

PR #81, opened 2017-04-14
Use controlled promotion of field inputs in constructors

### LINKED ISSUE #80: HSV(zero(N0f8), zero(N0f8), zero(N0f8)) causes a stack overflow
This is the actual stack overflow that occurs in https://github.com/JuliaImages/Images.jl/issues/624

To reproduce:

```julia
julia> using ColorTypes, FixedPointNumbers

julia> HSV(zero(N0f8), zero(N0f8), zero(N0f8))
ERROR: StackOverflowError:
 in ColorTypes.HSV{T<:AbstractFloat}(::FixedPointNumbers.Normed{UInt8,8}, ::FixedPointNumbers.Normed{UInt8,8}, ::FixedPointNumbers.Normed{UInt8,8}) at /Users/rdeits/.julia/v0.5/ColorTypes/src/types.jl:446 (repeats 80000 times)
```

This doesn't happen for the Gray or RGB constructors.

timholy:
Fixes #80 (CC @rdeits).

This also happens to fix a weird issue I've noticed for a long time: that the tests pass when you say `using ColorTypes` (which precompiles the module), but not if you say `include("../src/ColorTypes.jl"); include("runtests.jl")`. Satisfying to have that gone.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -443,7 +443,7 @@ macro make_constructors(C, fields, elty)
     esc(quote
         # More constructors for the non-alpha version
         $C{T<:Integer}($(Tfields...)) = $C{$elty}($(fields...))
-        $C($(realfields...)) = $C(promote($(fields...))...)
+        $C($(realfields...)) = $C{promote_eltype($C, $(fields...))}($(fields...))
         $C() = $C{$elty}($(zfields...))
     end)
 end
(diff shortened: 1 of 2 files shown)


timholy:
The test failure has to do with deprecations triggering warnings (or failing to do so depending on what's been called previously). Consequently it's not a reason to avoid merging this PR.

The deprecations at issue have been around for 6 months. Is that long enough to delete them?

rdeits:
Awesome, thanks!

tkelman:
What's the status on the deprecation test failure issue?

timholy:
Thanks for the reminder! See #82.

PR #82, opened 2017-04-29
Remove old deprecations

timholy:
See https://github.com/JuliaGraphics/ColorTypes.jl/pull/81#issuecomment-294139027. I added some tests to make sure the behavior we were trying to avoid doesn't get re-added.

### DIFF
--- a/src/ColorTypes.jl
+++ b/src/ColorTypes.jl
@@ -3,13 +3,7 @@ __precompile__()
 module ColorTypes
 
 using FixedPointNumbers
-if VERSION >= v"0.5.0-dev+755"
-    using Base.@pure
-else
-    macro pure(ex)
-        esc(ex)
-    end
-end
+using Base.@pure
 
 import Compat
 using Compat.TypeUtils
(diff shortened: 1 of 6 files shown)


tkelman:
Could do a grep through pkgeval logs to see who hasn't fixed these yet? what should be searched for? it's all public in the website repo https://github.com/JuliaCI/pkg.julialang.org but the abuse of git there means it's probably a huge repo to clone

timholy:
Good grep strings:
```
use reinterpret(RGB24
use reinterpret(Gray24
use reinterpret(ARGB32
use reinterpret(AGray32
use reinterpret(UInt32,x) == y
use x == reinterpret(UInt32,y)
```
I already have a fix for Images (https://github.com/JuliaImages/Images.jl/pull/634), so don't worry about that.

tkelman:
ImageMagick on 0.4 hits it, but apparently nothing else

timholy:
Since all recent versions of ImageMagick require julia 0.5, I think we're fine.

Issue #87, opened 2017-07-07
ColorTypes of Reals types, e.g. ForwardDiff.Dual


goretkin:
I was interested in taking automatic derivatives of `colordiff` with respect to RGB channels, but I cannot construct RGB Colors with any such elements that are subtypes of `Real`, because for all the concrete RGB types, the type parameter is `<:Fractional`

```julia
using Colors
using ForwardDiff

@show RGB(0.0, 0.0, ForwardDiff.Dual(0.5, 1.0))
@show Lab(0.0, 0.0, ForwardDiff.Dual(0.5, 1.0))
``` 

produces

```
RGB(0.0, 0.0, ForwardDiff.Dual(0.5, 1.0)) = RGB{N0f8}(0.0,0.0,0.502)
ERROR: MethodError: Cannot `convert` an object of type ForwardDiff.Dual{1,Float64} to an object of type Float32
This may have arisen from a call to the constructor Float32(...),
since type constructors fall back to convert methods.
Stacktrace:
 [1] ColorTypes.Lab(::Float64, ::Float64, ::ForwardDiff.Dual{1,Float64}) at /Users/goretkin/.julia/v0.6/ColorTypes/src/types.jl:442
```

The issue with Lab is different, because its type parameter is `<: AbstractFloat`.

Is there a downside to making all the type parameters `<: Real`?

SimonDanisch:
I'm quite sure that we can just change that restriction! Let's wait for
@timholy, but I think you can just make a pull request and see how it goes
:)

Am 07.07.2017 19:19 schrieb "Gustavo Goretkin" <notifications@github.com>:

timholy:
Not quite as obvious a decision as one might hope. The current system is designed to help "gently" enforce the 0-1 colorscale, meaning that it's possible to construct out-of-gamut values but we nevertheless have a catch for users who try this:
```julia
julia> using ColorTypes

julia> RGB(255, 0, 0)
ERROR: ArgumentError: (255, 0, 0) are integers in the range 0-255, but integer inputs are encoded with the N0f8
  type, an 8-bit type representing 256 discrete values between 0 and 1.
  Consider dividing your input values by 255, for example: RGB{N0f8}(255/255,0/255,0/255)
  See the READMEs for FixedPointNumbers and ColorTypes for more information.
Stacktrace:
[...]

PR #89, opened 2017-07-17
WIP: Use Real in place of AbstractFloat and Fractional

goretkin:
My goal is to do forward-mode automatic differentiation  through the colorimetry functions (see https://github.com/JuliaGraphics/ColorTypes.jl/issues/87).

Right now, I believe the element types for the colortypes are too restrictive. There's a good reason (https://github.com/JuliaGraphics/ColorTypes.jl/issues/87#issuecomment-313784538), but I'm hoping there's another way to accomplish that, and add it once we have generic type parameters.

With the current change (only changing the type definitions which use `<:AbstractFloat`), there's already trouble:

```julia
julia> convert(LMS, RGB(1,1,1))
ERROR: ArgumentError: FixedPointNumbers.Normed{UInt8,8} is an 8-bit type representing 256 values from 0.0 to 1.0; cannot represent 1.08883
Stacktrace:
 [1] throw_converterror(::Type{FixedPointNumbers.Normed{UInt8,8}}, ::Float64) at /Users/goretkin/.julia/v0.6/FixedPointNumbers/src/FixedPointNumbers.jl:170
 [2] _convert(::Type{FixedPointNumbers.Normed{UInt8,8}}, ::Type{UInt8}, ::Float64) at /Users/goretkin/.julia/v0.6/FixedPointNumbers/src/normed.jl:48
 [3] convert(::Type{ColorTypes.LMS}, ::ColorTypes.RGB{FixedPointNumbers.Normed{UInt8,8}}) at /Users/goretkin/.julia/v0.6/ColorTypes/src/conversions.jl:7
```

and the problem boils down to this commit causing:

```julia
julia> ColorTypes.issupported(LMS, eltype(RGB(1,1,1)))
true
```

whereas before it was `false`.

eltype(RGB(1,1,1)) is a subtype of `Real`, but it's not a subtype of `AbstractFloat`. If I follow the logic correctly, in the released version of `ColorTypes`, because `LMS` doesn't "support" the fixed-point type of `RGB(1,1,1)`, the values are promoted to `Float32` during conversion.

I'm looking for any suggestions for different logic to trigger this promotion. Probably any colorspace conversion from fixed point eltype (except perhaps to `Gray`, which is is a `<:Fractional` for whatever reason), should trigger a promotion to `Float32`.

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -161,7 +161,7 @@ end
 RGB4{T<:Fractional}(r::T, g::T, b::T) = RGB4{T}(r, g, b)
 
 "`HSV` is the Hue-Saturation-Value colorspace."
-immutable HSV{T<:AbstractFloat} <: Color{T,3}
+immutable HSV{T<:Real} <: Color{T,3}
     h::T # Hue in [0,360)
     s::T # Saturation in [0,1]
     v::T # Value in [0,1]
@@ -171,14 +171,14 @@ end
 const HSB = HSV
 
 "`HSL` is the Hue-Saturation-Lightness colorspace."
-immutable HSL{T<:AbstractFloat} <: Color{T,3}
+immutable HSL{T<:Real} <: Color{T,3}
     h::T # Hue in [0,360)
     s::T # Saturation in [0,1]
     l::T # Lightness in [0,1]
 end
 
 "`HSI` is the Hue-Saturation-Intensity colorspace."
-immutable HSI{T<:AbstractFloat} <: Color{T,3}
+immutable HSI{T<:Real} <: Color{T,3}
     h::T
     s::T
     i::T
@@ -189,63 +189,63 @@ end
 meaning that mathematical operations such as addition, subtraction,
 and scaling make "colorimetric sense" in this colorspace.
 """
-immutable XYZ{T<:AbstractFloat} <: Color{T,3}
+immutable XYZ{T<:Real} <: Color{T,3}
     x::T
     y::T
     z::T
 end
 
 "`xyY` is the CIE 1931 xyY (chromaticity + luminance) space"
-immutable xyY{T<:AbstractFloat} <: Color{T,3}
+immutable xyY{T<:Real} <: Color{T,3}
     x::T
     y::T
     Y::T
 end
 
 "`Lab` is the CIELAB colorspace."
-immutable Lab{T<:AbstractFloat} <: Color{T,3}
+immutable Lab{T<:Real} <: Color{T,3}
     l::T # Luminance in approximately [0,100]
     a::T # Red/Green
     b::T # Blue/Yellow
 end
 
 "`LCHab` is the Luminance-Chroma-Hue, Polar-Lab colorspace"
-immutable LCHab{T<:AbstractFloat} <: Color{T,3}
+immutable LCHab{T<:Real} <: Color{T,3}
     l::T # Luminance in [0,100]
     c::T # Chroma
     h::T # Hue in [0,360)
 end
 
 "`Luv` is the CIELUV colorspace"
-immutable Luv{T<:AbstractFloat} <: Color{T,3}
+immutable Luv{T<:Real} <: Color{T,3}
     l::T # Luminance
     u::T # Red/Green
     v::T # Blue/Yellow
 end
 
 "`LCHuv` is the Luminance-Chroma-Hue, Polar-Luv colorspace"
-immutable LCHuv{T<:AbstractFloat} <: Color{T,3}
+immutable LCHuv{T<:Real} <: Color{T,3}
     l::T # Luminance
     c::T # Chroma
     h::T # Hue
 end
 
 "`DIN99` is the (L99, a99, b99) adaptation of CIELAB"
-immutable DIN99{T<:AbstractFloat} <: Color{T,3}
+immutable DIN99{T<:Real} <: Color{T,3}
     l::T # L99
     a::T # a99
     b::T # b99
 end
 
 "`DIN99d` is the (L99d, a99d, b99d) improvement on DIN99"
-immutable DIN99d{T<:AbstractFloat} <: Color{T,3}
+immutable DIN99d{T<:Real} <: Color{T,3}
     l::T # L99d
     a::T # a99d
     b::T # b99d
 end
 
 "`DIN99o` is the (L99o, a99o, b99o) adaptation of CIELAB"
-immutable DIN99o{T<:AbstractFloat} <: Color{T,3}
+immutable DIN99o{T<:Real} <: Color{T,3}
     l::T # L99o
     a::T # a99o
     b::T # b99o
@@ -255,21 +255,21 @@ end
 `LMS` is the Long-Medium-Short colorspace based on activation of the
 three cone photoreceptors.  Like `XYZ`, this is a linear color space.
 """
-immutable LMS{T<:AbstractFloat} <: Color{T,3}
+immutable LMS{T<:Real} <: Color{T,3}
     l::T # Long
     m::T # Medium
     s::T # Short
 end
 
 "`YIQ` is a color encoding, for example used in NTSC transmission."
-immutable YIQ{T<:AbstractFloat} <: Color{T,3}
+immutable YIQ{T<:Real} <: Color{T,3}
     y::T
     i::T
     q::T
 end
 
 "`YCbCr` is the Y'CbCr color encoding often used in digital photography or video"
-immutable YCbCr{T<:AbstractFloat} <: Color{T,3}
+immutable YCbCr{T<:Real} <: Color{T,3}
     y::T
     cb::T
     cr::T


timholy:
Now that a lot more work has been done on the promotion and traits system here, want to try this again and see if you make more progress?

kimikage:
IMO, changing `Fractional` to `Real` is not much troublesome, but changing `AbstractFloat` to `Real` is.

Is it not enough to convert the constructor arguments or the struct fields?
When the colorspace is changed, many of the properties or characteristics of the original element type (e.g. `ForwardDiff`) will be lost. If you want to preserve those properties, we have to introduce a mechanism for lazy evaluation in color conversions.

goretkin:
> When the colorspace is changed, many of the properties or characteristics of the original element type (e.g. `ForwardDiff`) will be lost.

Sorry, I don't think I understand. On some level, changing the color space is just applying a function that maps 3 input numbers to 3 output numbers. It seems you can define color spaces and transformations between them without any need to specify the number representation type.

kimikage:
In general, supertypes cannot retain the full nature of subtypes. :confused:

For example, how can `RGB{Rational{Int}}` be converted to `XYZ{Rational{Int}}`?
Probably, it will be implemented as follows:
`RGB{Rational{Int}}`-->`RGB{Float64}`-->`XYZ{Float64}`-->`XYZ{Rational{Int}}`
I don't think there is a problem with such a conversion itself, but it definitely loses the properties of `Rational`.

I'm concerned about the lack of consensus, and I think there are technical solutions and compromises.

timholy:
I agree with the concerns, but I think being able to use automatic differentiation on colorant images is a big win; while the machine learning platforms will typically call the equivalent of `channelview`, there is a lot of merit in making the direct approach Just Work. (Esp. given our `ReinterpretArray` performance penalties.)

But I'm also not opposed to restricting allowed types more than `Real`, though. I've never really been comfortable with the fact that `Dual <: Real`, it's really not true. I understand, though, given that restriction in many packages and some missing elements to our numeric trait system.

kimikage:
Similarly, I strongly agree with the benefits of Real. But it brings many traps. Allowing `Real` means to allow (unsgined) `Normed`. As you know, `Lab` treats negative numbers. Of course there should be solutions. However, I hesitate to implement them.:sweat_smile:

goretkin:
My personal philosophy (that I do not hold with much conviction) is that types exist in julia in service of multiple dispatch. And not in service of really anything else.

For example, there's the opportunity to use types to encode non-negativity, but it's not done: `length` returns a signed integer, and not an unsigned integer.

So I think yes, allow `Real` (or heck even `Number`... it's so fuzzy what those interfaces actually are and what properties they're supposed to have), and rely on `InexactError`s if some conversion takes you out of the set of expressible numbers.

kimikage:
I think that the greatest outstanding feature in ColorTypes.jl is that we can use `RGB` instead of `RGB{T}` in the cases of conversions and so on.
It's easy to throw errors when problems with the element type occurs. However, it hurts the advantage significantly. @timholy et al. have made significant efforts for the advantage. I only helped with a bit about the promotion rules, I mean, just a bit, but I was very tired.:smile:

timholy:
@goretkin, as @kimikage says it's easy to underestimate the amount of work that has gone into the type system of this package. Click on the changes in recent PRs like #161, #169, #170 to get a taste for this. That said, I think if one can say `ccolor(RGB, Gray{ForwardDiff.Dual{ForwardDiff.Tag{typeof(f),Float64},Float64,3}})` and get the right answer (which might happen "automatically" given our current implementations), and also have a working `convert(C, x)` where `C` is that type, it would settle most of my concerns.

goretkin:
In case I've come off this way, I don't mean to diminish any of the work that has gone into this package. Thank you both for that!

I understand in this thread two recent concerns about `Real`: 
- it allows unsigned types (e.g. `UInt8`)  everywhere
- it allows fractional types (e.g. `Rational{Int64}`) everywhere

And this is a problem because unlike with `AbstractFloat`, many operations take you outside the values representable by the type. A`<:AbstractFloat` is going to be closed [1] under every color transformation. But those closure properties are not true for `UInt8` and `Rational{Int64}`.

Are these the traps you mean, @kimikage?

My response about philosophy is poorly worded, but I'm trying to allude to the general sense against overspecifiying/overconstraining (the word that has been used before is escaping me) types annotations. That applies to method definitions, but indirectly to type definitions if types are the only way to access the inputs/outputs of a mathematical function.

I might have misunderstood, and I heard a different point about leveraging the type system to enforce constraints, such as non-negativity. That's why I mentioned `length` (though a better example would be if `getindex(::Vector, i::Unsigned)` were defined in place of a more general constraint on `i`, and you could not index with a `Signed`)

I sense I'm missing some deeper point here. I agree that it's a great feature, worthy of preservation, to be able to name conversions into color spaces without naming the parameterization of the type (`RGB` instead of `RGB{N0f8}`. I didn't understand how allowing construction of an `RGB{<:Real}` interferes with that.

Unfortunately, I haven't put the necessary time into understanding/remembering even what the original issue I was trying to solve was, and I likely won't be able to in the near future. So at the very least I'd like to make sure I'm not getting in the way of the development of this package :-D

[1] in the sense of https://en.wikipedia.org/wiki/Closure_(mathematics)

timholy:
Just to clarify, in no way did I feel unappreciated. Just wanting to make sure you understood the constraints!

cdsousa:
Hi all, is the problem present in the folloing code related to this issue?

```julia
using StaticArrays, ColorTypes
import ForwardDiff, ForwardDiff2, Zygote

ForwardDiff.gradient(x -> RGB(x[1], 0.5, 0.5).r, SVector(0.0))[1] ########## Wrong: returns 0.0N0f8
ForwardDiff2.DI(x -> RGB(x, 0.5, 0.5).r)(0.0) ########## Wrong: returns 0.00

Zygote.gradient(x -> RGB(x, 0.5, 0.5).r, 0.0) ########## Correct: returns 1.00
struct C{T}
    a::T
    b::T
end
ForwardDiff.gradient(x -> C(x[1], x[1]).a, SVector(0.0))[1] ########## Correct: returns 1.00
ForwardDiff2.DI(x -> C(x, x).a)(0.0) ########## Correct: returns 1.00
```

kimikage:
> is the problem present in the following code related to this issue?

I guess so, but I think being able to calculate gradients correctly can be considered separately from extending the element type to `Real`.

```julia
julia> ForwardDiff.gradient(x -> RGB{Float64}(x[1], 0.5, 0.5).r, SVector(0.0))[1]
ERROR: MethodError: no method matching Float64(::ForwardDiff.Dual{ForwardDiff.Tag{var"#29#30",Float64},Float64,1})

julia> ForwardDiff.gradient(x -> Float64(x[1]), SVector(0.0))[1]
ERROR: MethodError: no method matching Float64(::ForwardDiff.Dual{ForwardDiff.Tag{var"#31#32",Float64},Float64,1})
```

kimikage:
The case study with the `RGB{Rational{Int}}`->`XYZ{Rational{Int}}` conversion (cf. https://github.com/JuliaGraphics/ColorTypes.jl/pull/89#issuecomment-591771457) implies that we cannot treat `Dual` as `Real` without specialization. In fact, "mathematically", a dual number is not a real number due to the `ϵ`.

If we do the specialization to handle `Dual`, I think it should be done in a new package, like `ColorForwardDiff.jl`.

cdsousa:
Can I open an issue about making ColorTypes forward diff / dual friendly?

cdsousa:
Or maybe this is something that can be workable from ForwardDiff2 with the source-to-source techniques?

kimikage:
Since the `RGB` space is usually expressed as a orthogonal coordinate system, I don't think it is necessary to use the `RGB` type instead of generic vectors. Therefore, I think it is important where we set our goal. The derivative of `colordiff` with `RGB` inputs is an example of a symbolic goal, but I don't know the benefits of not using Zygote.jl.

goretkin:
Note that forward diff is one motivating example, but there are others: https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/478d673b61e1c7883bb6c05d08b72c53ada18dd7/src/intervals/intervals.jl#L14-L36

A solution that involves code specifically to handle `ColorTypes.jl` and `ForwardDiff.jl` seems to go against one of the main emergent themes of Julia, which is that packages that don't know about each other can still work together.

kimikage:
`Interval` is not a "real" number, too! :sweat_smile: Why is such a lazy evaluation necessary for color types? Some constants used in Colors.jl have only 3-4 significant digits.:confused:

goretkin:
> Interval is not a "real" number, too! 
All sorts of weird type relations happen in julia libraries because

1. it's not really clear what a `Number`, `Real` or `AbstractFloat` really are. 
2. they want to use things like `Colors.RGB` and `Complex`, which constrain the type parameter to e.g. `AbstractFloat` or `Real`.

> Why is such a lazy evaluation necessary for color types? 
You could use an interval for lots of reasons, in addition to handling rounding errors. For example: https://en.wikipedia.org/wiki/MacAdam_ellipse

Also, even if we can't think of a reason now, it doesn't mean one doesn't exist. Again this is the beauty of a lot of the Julia ecosystem. Things Just Work more often than in other languages :-D.

kimikage:
So, do you think it's better to make it `Any`, not `Real`? :stuck_out_tongue:
I have a bad feeling about the relaxation, even though I think the element type should be "mathematically" real numbers.

goretkin:
> So, do you think it's better to make it `Any`, not `Real`? 😛

(a year later to the day!)

Probably, yes. In my own experience, I have usually regretted it when I (or someone else) put type constraints in structs.
I doubt there is a perfect type in the hierarchy to capture the interface that `ColorTypes.jl` requires of the element type, without imposing any additional constraints.

The interface used, I'm guessing, at least a [field](https://en.wikipedia.org/wiki/Field_(mathematics)) (but even that's not entirely accurate, because IEEE floats aren't associative). I don't know how to trace all of the methods that are called on an element type when e.g. running this package's tests, but it would be very cool and helpful to see that (in general!).

kimikage:
Now that the package extension mechanism has been introduced, if the handling of special types can be done outside of ColorTypes.jl and Colors.jl, 80% of my concerns will be resolved.

However, we still have not reached a conclusion on how to determine the destination component type in the color conversion. Also, the implementation is incomplete.

One possible policy is to match the default component type to the constructor's decision.
It is clearly technically possible; it is just a matter of who will do the work.

PR #90, opened 2017-07-25
fix depwarns under 0.7 and bump REQUIRE and CI to 0.6

Sacha0:
Best! :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
   - linux
   - osx
 julia:
-  - 0.5
   - 0.6
   - nightly
 notifications:
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
-julia 0.5
+julia 0.6
 Compat 0.26.0
 FixedPointNumbers 0.3.0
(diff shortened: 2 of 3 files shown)


timholy:
Sure, seems like time to move on. Thanks!

lobingera:
and 0.5 users?

ararslan:
People on 0.5 can continue to use the last tagged version that supports 0.5. If something needs fixing, you can branch from the last 0.5 tag, cherry-pick commits from master, and do another 0.5-supporting release.

Sacha0:
Thanks all! :)

lobingera:
@ararslan Maybe i'm a little bit over sensitive on this, but as ColorTypes is a requirement for Colors and this in turn is a requirement for almost everything that plots/draws/paints something; isn't this one of the cases where a simple Pkg.update() in 0.5 can filp you whole package repository to unusable?

timholy:
No, since you can't install the 0.6.x series of ColorTypes, you'll continue to use the 0.5.x series as long as you're relying on `Pkg.update()` and not following `master`. (Use `Pkg.free()` if necessary.)

Issue #84, opened 2017-05-17
gamut of Lab is not consitent with behavour of convert


oxinabox:
apparently the gamut of Lab (Hunter, 1971 iirc) is:

```julia
julia>gamutmax(Lab)
(100,1,1)

julia>gamutmin(Lab)
(0,0,0)
```

Let us check that, by converting a color (using Colors.jl)

```julia
julia> repr( convert(Lab, RGB(0.1,0.9,0.2)))
"Lab{Float64}(80.04670273385801, -76.74357468570415, 68.26355168645148)"
```

A perfectly nice green, as expected: `#1AE633`
But out of the `gamut` given.

------------

By checking a large number of colors (the train portion of the Monroe color dataset: 1_523_108 items) and getting the extrema this is what i found:

---
`#FFFFFF`
```
eg_Lab = Lab{Float32}(100.00001f0,-2.5005475f-5,8.302305f-6)
convert(RGB,eg_Lab) = RGB{Float32}(1.0f0,1.0f0,1.0f0)
```

---
`#FF04FF`
```
eg_Lab = Lab{Float32}(60.401684f0,98.00463f0,-60.70089f0)
convert(RGB,eg_Lab) = RGB{Float32}(1.0f0,0.015686838f0,1.0f0)
```

---
`#FDFE02`
```
eg_Lab = Lab{Float32}(96.72747f0,-21.885283f0,93.968506f0)
convert(RGB,eg_Lab) = RGB{Float32}(0.99215686f0,0.99607855f0,0.007842416f0)
```

---
`#020000`
```
eg_Lab = Lab{Float32}(0.11661911f0,0.5230051f0,0.18427981f0)
convert(RGB,eg_Lab) = RGB{Float32}(0.007843167f0,0.0f0,3.6899073f-9)
```

---
`#01FF05`
```
eg_Lab = Lab{Float32}(87.74314f0,-86.084785f0,82.80794f0)
convert(RGB,eg_Lab) = RGB{Float32}(0.0039214566f0,1.0f0,0.019607343f0)
```
---
`#0301FF`
```
eg_Lab = Lab{Float32}(32.38845f0,79.05015f0,-107.70611f0)
convert(RGB,eg_Lab) = RGB{Float32}(0.011765122f0,0.0039214636f0,1.0f0)

```

In those observations we saw a
```julia
julia> minimum(getfield.(all_examples_Lab, fieldnames(Lab)'),1)

[...]

PR #94, opened 2017-08-16
Fix Lab gamut (#84)

oxinabox:
This just adds the conclusions from the discussion of #84.

Can anyone think of a good set of tests?


timholy:
Other than directly testing the same values entered in source, the only "consumer" is `rand`. So the only easy way to test would be to generate lots of random numbers. We could conceivably add a test to see how much of RGB space gets filled by `[convert(RGB, rand(C)) for i = 1:10^4]` for all the `C` types, but that's not entirely easy.

Any ideas, @bjarthur?

bjarthur:
not really.  perhaps @m-lohmann has some?

timholy:
I'm inclined to merge this without tests, just to avoid causing @oxinabox any hassle from merge conflicts with #92. Seem OK?

bjarthur:
fine with me

timholy:
Thanks, @oxinabox!

PR #96, opened 2017-09-18
Switch to oneunit where appropriate

timholy:
This also modifies the implementation of a non-exported function for use in Colors


timholy:
Curious failure on nightly (linux only), but I notice the linux nightly is 4 days older than the mac one (which passed). I just built a recent julia master on my own machine and I can't replicate the error. So I'm going to assume this was a transient issue already fixed in base julia.

PR #97, opened 2017-09-20
Fix 0.7 depwarns

yuyichao:


timholy:
In case you didn't see it, https://github.com/JuliaGraphics/ColorTypes.jl/pull/96#issuecomment-330537320 with regards to the failure.

PR #103, opened 2018-01-31
Use isconcetetype and isabstracttype to unbreak on 0.7

andreasnoack:
Still a lot of deprecations though but this should make the tests pass (except for upstream issues in `FixedPointArrays`).


test/runtests.jl
```diff
@@ -1,7 +1,9 @@
 using ColorTypes, FixedPointNumbers
 using Base.Test
 
-@test isempty(detect_ambiguities(ColorTypes, Base, Core))
+if VERSION < v"0.7.0-"
```
timholy: Why conditional? Are there new ambiguities on 0.7?

andreasnoack: Yes. The following are now detected as ambiguous but not triggered by any of the tests
```
julia> detect_ambiguities(ColorTypes, Base, Core)
6-element Array{Tuple{Method,Method},1}:
 ((::Type{FixedPointNumbers.Fixed{T,f}})(x) where {T, f} in FixedPointNumbers at /Users/andreasnoack/.julia/v0.7/FixedPointNumbers/src/fixed.jl:8, (::Type{T})(z::Complex) where T<:Real in Base at complex.jl:37)
 ((::Type{FixedPointNumbers.Fixed{T,f}})(x) where {T, f} in FixedPointNumbers at /Users/andreasnoack/.julia/v0.7/FixedPointNumbers/src/fixed.jl:8, (::Type{T})(x::Char) where T<:Number in Core at boot.jl:681)
 ((::Type{FixedPointNumbers.Normed{T,f}})(x) where {T, f} in FixedPointNumbers at /Users/andreasnoack/.julia/v0.7/FixedPointNumbers/src/normed.jl:8, (::Type{T})(z::Complex) where T<:Real in Base at complex.jl:37)
 ((::Type{FixedPointNumbers.Normed{T,f}})(x) where {T, f} in FixedPointNumbers at /Users/andreasnoack/.julia/v0.7/FixedPointNumbers/src/normed.jl:8, (::Type{T})(x::Char) where T<:Number in Core at boot.jl:681)
 ((::Type{FixedPointNumbers.Fixed{T,f}})(x) where {T, f} in FixedPointNumbers at /Users/andreasnoack/.julia/v0.7/FixedPointNumbers/src/fixed.jl:8, (::Type{T})(x::Base.TwicePrecision) where T<:Number in Base at twiceprecision.jl:243)
 ((::Type{FixedPointNumbers.Normed{T,f}})(x) where {T, f} in FixedPointNumbers at /Users/andreasnoack/.julia/v0.7/FixedPointNumbers/src/normed.jl:8, (::Type{T})(x::Base.TwicePrecision) where T<:Number in Base at twiceprecision.jl:243)
```
Would be good to have them resolved eventually but thought it wasn't important enough to hold back the other fixed (I'm working myself though Gadfly's dependencies to make it work on 0.7)

lobingera:
https://github.com/JuliaMath/FixedPointNumbers.jl/issues/101

(and as you can get, this was a showstopper for Cairo.jl for some time).
@andreasnoack with the mentioned merge of 100 and some local code Cairo is testing successfully on 0.7dev

lobingera:
but now i'm confused, this is merged and succeeds on 0.7?

test/runtests.jl
timholy: Hmm, I didn't read the other one carefully enough. I'm against a cascading removal of the ambiguity tests, I'd have to remove them in about a dozen packages in the Images world. Better to fix them at the source.

andreasnoack: I'm probably influenced by my distaste for testing for ambiguities like this. If an ambiguity can't be reached by the normal tests then I don't think it is relevant. However, this package is your domain so I'll fix these if you give a hint to how you want them fixed. Should the constructor of a `Fixed` throw when fed a `Char`, `Complex`, or 'TwicePrecision` or should I convert via `Float64`?

timholy: Your point is a good one, and indeed one I used to adopt myself. But in the fullness of time I got enough bug reports from ambiguities *not* caught by the existing test suite that I've switched my view. I don't really care about ambiguities between packages, but my experience is that ambiguities with Base are generally worth fixing. In this situation, the `Char` ambiguity is indeed silly, and if there were 100 others like it I'd change my tune about this.

For `x::Complex{T}` I'd `convert(T, x)`; for `TwicePrecision` I'd convert via `Float64`, and for `Char` I'd throw an error.

andreasnoack:
Okay. This one should also be good to go now. I've reenabled the ambiguity tests and also squashed most of the deprecation warnings.

andreasnoack:
Thanks. Would it be possible to tag a new release?

PR #111, opened 2018-08-16
Export types used in other packages of the Images.jl project

juliohm:
This PR exports missing types used in other packages of the Images.jl project. I appreciate if you can review and merge. I will submit another PR to Colors.jl and Images.jl meanwhile.

### DIFF
--- a/src/ColorTypes.jl
+++ b/src/ColorTypes.jl
@@ -14,7 +14,7 @@ export Fractional
 
 export Colorant
 export Color, TransparentColor, AlphaColor, ColorAlpha, AbstractRGB
-export AbstractGray, TransparentGray, TransparentRGB
+export AbstractGray, Color3, TransparentGray, Transparent3, TransparentRGB, ColorantNormed
 
 export RGB, BGR, RGB1, RGB4
 export HSV, HSB, HSL, HSI
--- a/src/ColorTypes.jl
+++ b/src/ColorTypes.jl
@@ -14,6 +14,7 @@ export Fractional
 
 export Colorant
 export Color, TransparentColor, AlphaColor, ColorAlpha, AbstractRGB
+export AbstractGray, TransparentGray, TransparentRGB
 
 export RGB, BGR, RGB1, RGB4
 export HSV, HSB, HSL, HSI
--- a/src/types.jl
+++ b/src/types.jl
@@ -64,7 +64,7 @@ alpha channel comes last in the internal storage order.
 """
 abstract type ColorAlpha{C,T,N} <: TransparentColor{C,T,N} end
 
-# These are types we'll dispatch on. Not exported.
+# These are types we'll dispatch on.
 AbstractGray{T}                    = Color{T,1}
 Color3{T}                          = Color{T,3}
 TransparentGray{C<:AbstractGray,T} = TransparentColor{C,T,2}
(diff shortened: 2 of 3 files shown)


Evizero:
xref: https://github.com/JuliaImages/Images.jl/issues/731

timholy:
Ideally one would add a test that these exports are available, but if you're OK risking that someone else will come along and unexport them, we could merge this.

juliohm:
That is a great idea, how to implement such test? What function in Julia can be used to check for names exported?

timholy:
Many possibiliites. You could use `isa`, but I kind of like
```julia
dispatcher(::AbstractGray) = 1
dispatcher(::TransparentGray) = 2
...
@test dispatcher(Gray(0.2)) == 1
@test dispatcher(AGray(0.2, 0.5)) == 2
...
```

juliohm:
@timholy added tests to all types except for `Transparent3` and `ColorantNormed`. Could you please clarify which concrete types implement the concept?

timholy:
`ColorantNormed{T<:Normed,N} = Colorant{T,N}` just means any Colorant using Normed numbers (e.g., `N0f8`) as its eltype. 

`Transparent3{C<:Color3,T} = TransparentColor{C,T,4}` means a colorant that has transparency and its base color is a `Color3`. A good example is `RGBA`.

juliohm:
@timholy `RGBA` dispatches to `TransparentRGB` instead of `Transparent3`. Is there another concrete type that is not RGB that I could use in the test?

juliohm:
Also, how we can test the `ColorantNormed` case?

Evizero:
> Also, how we can test the ColorantNormed case?

 maybe something like

```julia
julia> using ColorTypes, FixedPointNumbers

julia> RGB{N0f8}(0,0.1, 0.5)
RGB{N0f8}(0.0,0.102,0.502)

julia> dispatcher(::ColorTypes.ColorantNormed) = 3
dispatcher (generic function with 1 method)

julia> dispatcher(::ColorTypes.Colorant) = 4
dispatcher (generic function with 2 methods)

julia> dispatcher(RGB{N0f8}(0.,0.1,0.5))
3

julia> dispatcher(RGB{Float32}(0.,0.1,0.5))
4
```

Evizero:
> `RGBA` dispatches to `TransparentRGB` instead of `Transparent3`. Is there another concrete type that is not RGB that I could use in the test?

found one:

```julia
julia> alphacolor(rand(HSV))
AHSV{Float64}(354.887716223749,0.259748993210698,0.7149158244136269,1.0)

julia> alphacolor(rand(HSV)) isa ColorTypes.TransparentRGB
false

julia> alphacolor(rand(HSV)) isa ColorTypes.Transparent3
true
```

juliohm:
Thank you @Evizero , for the `ColorantNormed` example you shared, it is dispatching on `Color3` with higher priority. Do you suggest another type?

juliohm:
Maybe the `ColorantNormed` doesn't need to exist? The dispatcher seems to always give priority to other types. I appreciate if you can clarify the issue.

Evizero:
I don't have a good answer to that.

Maybe the following test is good enough

```julia
@test RGB{N0f8} <: ColorantNormed
@test !(RGB{Float32} <: ColorantNormed)
```

timholy:
Or use a new function
```julia
normeddispatcher(::ColorantNormed) = 1
normeddispatcher(::Colorant) = 2
```

juliohm:
Awesome! The PR should be ready to merge. Thanks for all the suggestions.

timholy:
Thank you!

PR #116, opened 2018-12-15
[RFC] RGB16

### LINKED ISSUE #114: Implement RGB16 / RGB565
Hello,

following https://github.com/JuliaGraphics/Colors.jl/issues/332 implementing a 2 bytes (16 bits) colors might be considered

Here is a schema for such colors representation
```
1        |1          |        0
5 4 3 2 1|0 9 8 7 6 5|4 3 2 1 0 
---------|-----------|---------
R R R R R|G G G G G G|B B B B B
```

- 5 bits for **red** channel - 32 values

- 6 bis for **green** channel - 64 values (green gets an extra bit because it's the main element of luminance)

- 5 bits for **blue** channel - 32 values

Name|R|G|B|value
----|-|-|-|-----
BLACK|0|0|0|0
WHITE|31|63|31|65535
RED|31|0|0|63488
GREEN|0|63|0|2016
BLUE|0|0|31|31
GRAY|16|33|16|33840
BROWN|23|34|0|48192
YELLOW|31|63|0|65504

Kind regards

scls19fr:
Closes #114

### DIFF
--- a/src/ColorTypes.jl
+++ b/src/ColorTypes.jl
@@ -25,7 +25,7 @@ export YIQ, YCbCr
 export Gray
 
 export RGB24, ARGB32, Gray24, AGray32
-export RGB16
+export RGB565
 
 # Note: the parametric TransparentColorColors are exported
 # algorithmically, see `@make_alpha` in types.jl.
--- a/src/ColorTypes.jl
+++ b/src/ColorTypes.jl
@@ -25,6 +25,7 @@ export YIQ, YCbCr
 export Gray
 
 export RGB24, ARGB32, Gray24, AGray32
+export RGB16
 
 # Note: the parametric TransparentColorColors are exported
 # algorithmically, see `@make_alpha` in types.jl.
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -15,21 +15,21 @@ red(c::AbstractRGB   ) = c.r
 red(c::TransparentRGB) = c.r
 red(c::RGB24)  = N0f8((c.color & 0x00ff0000)>>16, 0)
 red(c::ARGB32) = N0f8((c.color & 0x00ff0000)>>16, 0)
-red(c::RGB16) = (c.value & 0xf800) >> 11
+red(c::RGB16) = reinterpret(N3f5, (c.value & 0xf800) >> 11)
 
 "`green(c)` returns the green component of an `AbstractRGB` opaque or transparent color."
 green(c::AbstractRGB   ) = c.g
 green(c::TransparentRGB) = c.g
 green(c::RGB24)  = N0f8((c.color & 0x0000ff00)>>8, 0)
 green(c::ARGB32) = N0f8((c.color & 0x0000ff00)>>8, 0)
-green(c::RGB16) = (c.value & 0x07e0) >> 5
+green(c::RGB16) = reinterpret(N2f6, (c.value & 0x07e0) >> 5)
 
 "`blue(c)` returns the blue component of an `AbstractRGB` opaque or transparent color."
 blue(c::AbstractRGB   ) = c.b
 blue(c::TransparentRGB) = c.b
 blue(c::RGB24)  = N0f8(c.color & 0x000000ff, 0)
 blue(c::ARGB32) = N0f8(c.color & 0x000000ff, 0)
-blue(c::RGB16) = c.value & 0x001f
+blue(c::RGB16) = reinterpret(N3f5, c.value & 0x001f)
 
 "`gray(c)` returns the gray component of a grayscale opaque or transparent color."
 gray(c::Gray)    = c.val
@@ -304,6 +304,7 @@ for T in (RGB24, ARGB32, Gray24, AGray32)
         reinterpret(::Type{$T}, x::UInt32) = $T(x, Val{true})
     end
 end
+reinterpret(::Type{UInt16}, c::RGB16) = c.value
 ==(x::AbstractGray, y::AbstractGray) = gray(x) == gray(y)
 ==(x::Number, y::AbstractGray) = x == gray(y)
 ==(x::AbstractGray, y::Number) = ==(y, x)
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -15,18 +15,21 @@ red(c::AbstractRGB   ) = c.r
 red(c::TransparentRGB) = c.r
 red(c::RGB24)  = N0f8((c.color & 0x00ff0000)>>16, 0)
 red(c::ARGB32) = N0f8((c.color & 0x00ff0000)>>16, 0)
+red(c::RGB16) = (c.value & 0xf800) >> 11
 
 "`green(c)` returns the green component of an `AbstractRGB` opaque or transparent color."
 green(c::AbstractRGB   ) = c.g
 green(c::TransparentRGB) = c.g
 green(c::RGB24)  = N0f8((c.color & 0x0000ff00)>>8, 0)
 green(c::ARGB32) = N0f8((c.color & 0x0000ff00)>>8, 0)
+green(c::RGB16) = (c.value & 0x07e0) >> 5
 
 "`blue(c)` returns the blue component of an `AbstractRGB` opaque or transparent color."
 blue(c::AbstractRGB   ) = c.b
 blue(c::TransparentRGB) = c.b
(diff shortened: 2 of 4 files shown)


src/types.jl
```diff
@@ -160,6 +160,23 @@ struct RGB4{T<:Fractional} <: AbstractRGB{T}
...
 RGB4(r::T, g::T, b::T) where {T<:Fractional} = RGB4{T}(r, g, b)
 
+"""`RGB16` is Red-Green-Blue colorspace represented by two bytes (16 bits).
+
+1        |1          |        0
+5 4 3 2 1|0 9 8 7 6 5|4 3 2 1 0 
+---------|-----------|---------
+R R R R R|G G G G G G|B B B B B
+"""
+struct RGB16
```
timholy: Should be `<: AbstractRGB`

src/traits.jl
```diff
@@ -15,18 +15,21 @@ red(c::AbstractRGB   ) = c.r
 red(c::TransparentRGB) = c.r
 red(c::RGB24)  = N0f8((c.color & 0x00ff0000)>>16, 0)
 red(c::ARGB32) = N0f8((c.color & 0x00ff0000)>>16, 0)
+red(c::RGB16) = (c.value & 0xf800) >> 11
```
timholy: Consider `reinterpret(N3f5, (c.value & 0xf800) >> 11)`.

src/traits.jl
```diff
@@ -15,18 +15,21 @@ red(c::AbstractRGB   ) = c.r
...
 
 "`green(c)` returns the green component of an `AbstractRGB` opaque or transparent color."
 green(c::AbstractRGB   ) = c.g
 green(c::TransparentRGB) = c.g
 green(c::RGB24)  = N0f8((c.color & 0x0000ff00)>>8, 0)
 green(c::ARGB32) = N0f8((c.color & 0x0000ff00)>>8, 0)
+green(c::RGB16) = (c.value & 0x07e0) >> 5
```
timholy: This should be an `N2f6`

src/traits.jl
```diff
@@ -15,18 +15,21 @@ red(c::AbstractRGB   ) = c.r
...
 
 "`blue(c)` returns the blue component of an `AbstractRGB` opaque or transparent color."
 blue(c::AbstractRGB   ) = c.b
 blue(c::TransparentRGB) = c.b
 blue(c::RGB24)  = N0f8(c.color & 0x000000ff, 0)
 blue(c::ARGB32) = N0f8(c.color & 0x000000ff, 0)
+blue(c::RGB16) = c.value & 0x001f
```
timholy: Another `N3f5`

src/types.jl
```diff
@@ -160,6 +160,23 @@ struct RGB4{T<:Fractional} <: AbstractRGB{T}
...
+    value::UInt16
+end
+
+function RGB16(r, g, b)
+    RGB16(r << 11 + g << 5 + b)
```
timholy: This implementation makes strong assumptions about the input types and formats. I would restrict the input arguments to `N3f5`/`N2f6` as appropriate, and then introduce "conversion" constructors for other input types.

src/types.jl
```diff
@@ -160,6 +160,23 @@ struct RGB4{T<:Fractional} <: AbstractRGB{T}
...
+end
+
+UInt16(c::RGB16) = c.value
```
timholy: `UInt16(x)` converts a number while preserving its value, but here you're thinking of this as a bit pattern. Consequently this should be `reinterpret(::Type{UInt16}, c::RGB16) = c.value`.

src/types.jl
scls19fr: I'm getting `LoadError: invalid subtyping in definition of RGB16`

timholy: Ah, OK, this is a little more complicated. I think you should make it `RGB16 <: AbstractRGB{N2f6}` and then make sure that `red`, `green`, and `blue` all return an `N2f6`. This requires small changes in your mask and shifting.

Alternatively, `RGB16 <: AbstractRGB{Union{N3f5,N2f6}}` *seems* to work, though I wonder if that might cause trouble elsewhere. You could try it and see; for example, if you create an `RGB16` array, in theory ImageView.jl should just display it without complaint. If you encounter errors, this gives you good guidance about what to fix.

src/types.jl
timholy: Also note that since

```julia
julia> typemax(N2f6)
4.05N2f6

julia> typemax(N3f5)
8.23N3f5
```

you might need to add some checks. (Both of these correspond to the bit pattern `0xff`. This is equivalent to making sure that `r`, `g`, and `b` don't have bits set that you're not expecting.)

scls19fr: I won't be able to do more during the next 2 coming weeks... 
If someone here could take the lead on this that will be very nice.

timholy: No worries, I don't think anyone else is asking for this. It can wait until you have time.

src/ColorTypes.jl
```diff
@@ -25,6 +25,7 @@ export YIQ, YCbCr
 export Gray
 
 export RGB24, ARGB32, Gray24, AGray32
+export RGB16
```
timholy: Rename to RGB565 as discussed

src/types.jl
```diff
@@ -160,6 +160,21 @@ struct RGB4{T<:Fractional} <: AbstractRGB{T}
...
 RGB4(r::T, g::T, b::T) where {T<:Fractional} = RGB4{T}(r, g, b)
 
+"""`RGB565` is Red-Green-Blue colorspace represented by two bytes (16 bits).
+
+1        |1          |        0
+5 4 3 2 1|0 9 8 7 6 5|4 3 2 1 0 
+---------|-----------|---------
+R R R R R|G G G G G G|B B B B B
+"""
+struct RGB565 <: AbstractRGB
```
timholy:
```suggestion
struct RGB565 <: AbstractRGB{Normed}
```

Or make it `AbstractRGB{N2f6}` and use `N2f6` throughout (in both constructors and `red`/`green`/`blue`).

src/types.jl
```diff
@@ -160,6 +160,21 @@ struct RGB4{T<:Fractional} <: AbstractRGB{T}
...
+    value::UInt16
+end
+
+function RGB565(r, g, b)
```
timholy: You could call this `RGB565_integer` or something, but the default constructor has to respect the 0-to-1 colorscale (essentially being the converse of `red`, `green`, `blue`). Otherwise you get into a situation where `1` has a different *mathematical* (rather than computational) meaning than `1.0`, which would be a disaster.

`RGB565` itself has to be something like

```julia
function RGB565(r::N3f5, g::N2f6, b::N3f5)
    # body. Needs to check that all inputs are between 0 and 1.
end
RGB565(r, g, b) = RGB565(N3f5(r), N2f6(g), N3f5(b))
```

test/runtests.jl
```diff
@@ -659,4 +659,33 @@ end
     @test RGB(Gray(0.2), Gray(0.3), Gray(0.4)) == RGB(0.2, 0.3, 0.4)
 end
 
+@testset "RGB565" begin
+    c = RGB565(31, 63, 31)  # white
```
timholy:
```suggestion
    c = RGB565(1, 1, 1)  # white
```

Isn't that much more clearly white? (If you prefer you could write this `RGB565(1.0, 1.0, 1.0)`.)

test/runtests.jl
```diff
@@ -659,4 +659,33 @@ end
...
+    @test blue(c) == 31
```
timholy:
```suggestion
    @test blue(c) === N3f5(1)
```

Similar changes throughout below.

scls19fr:
I let other contributor having a look at this.
Sorry.

test/runtests.jl
```diff
@@ -659,4 +659,33 @@ end
...
+    @test green(c) == 63
+    @test red(c) == 31
+    @test UInt16(c) == 65535
```
timholy:
```suggestion
    @test reinterpret(UInt16, c) == 65535
```
If `UInt16(c)` does the same thing, that is a serious bug and needs to be fixed.

timholy:
OK. I don't plan on adding it myself, so it will rely on someone being sufficiently enthusiastic to put in the effort.

PR #120, opened 2019-05-13
add Project.toml and drop 0.7 support

johnnychen94:
it's time to move to the new Registrator 😄


johnnychen94:
Seems that I should have bumped minor version instead of patch version. But I’m off from my computer now..

PR #121, opened 2019-05-13
expand floattype from FixedPointNumbers

johnnychen94:
This patch is a generalization of https://github.com/JuliaImages/Images.jl/issues/803#issuecomment-488139972

Expanding the existing `FixedPointNumbers.floattype` seems reasonable to me, but if there are concerns on type piracy, I'm also happy to use a new name.

Blocking PR:

- [x] https://github.com/JuliaMath/FixedPointNumbers.jl/pull/122

### DIFF
--- a/src/ColorTypes.jl
+++ b/src/ColorTypes.jl
@@ -1,6 +1,7 @@
 module ColorTypes
 
 using FixedPointNumbers
+import FixedPointNumbers: floattype
 using Base: @pure
 
 const Fractional = Union{AbstractFloat, FixedPoint}
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -172,6 +172,26 @@ and the easiest to use:
 """
 base_colorant_type(c::Colorant) = base_colorant_type(typeof(c))
 
+"""
+    floattype(::Type{T}) where T<:Colorant
+
+Promote storage data type of colorant `T` to `AbstractFloat` while keep the
+`base_colorant_type` the same.
+
+!!! info
+
+    Non-parametric colorants will be promote to corresponding parametric
+    colorants. For example, `floattype(RGB24) == RGB{Float32}`
+"""
+floattype(::Type{T}) where T <: Colorant =
+    base_colorant_type(T){floattype(eltype(T))} # 1 parameter
+# 0 parameter
+floattype(::Type{RGB24}) = RGB{Float32}
+floattype(::Type{Gray24}) = Gray{Float32}
+floattype(::Type{ARGB32}) = ARGB{Float32}
+floattype(::Type{AGray32}) = AGray{Float32}
+
+
 colorant_string(::Type{Union{}}) = "Union{}"
 colorant_string(::Type{C}) where {C<:Colorant} = string(nameof(C))
 function colorant_string_with_eltype(::Type{C}) where {C<:Colorant}
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,6 +1,8 @@
 using ColorTypes, FixedPointNumbers
 using Test
 
+@testset "ColorTypes" begin
+
 @test isempty(detect_ambiguities(ColorTypes, Base, Core))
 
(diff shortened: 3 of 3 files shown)


src/ColorTypes.jl
```diff
@@ -31,7 +32,7 @@ export RGB24, ARGB32, Gray24, AGray32
 
 
 ## Functions
-export base_color_type, base_colorant_type, ccolor, color, color_type
+export base_color_type, base_colorant_type, floattype, ccolor, color, color_type
```
timholy: FixedPointNumbers doesn't export `floattype`, this should be consistent.

src/traits.jl
```diff
@@ -172,6 +172,31 @@ and the easiest to use:
...
+
+```julia
+julia> T = RGB{N0f8}
+RGB{Normed{UInt8,8}}
+
+julia> base_colorant_type(T){floattype(T)}
+RGB{Float32}
+```
+"""
+floattype(::Type{T}) where T <: AbstractFloat = T
```
timholy: It's type-piracy to do this here, but wouldn't be in FixedPointNumbers

src/traits.jl
```diff
@@ -172,6 +172,31 @@ and the easiest to use:
...
+floattype(::Type{T}) where T <: Bool = Float32
+floattype(::Type{T}) where T <: Colorant = floattype(eltype(T))
```
timholy: This is *not* piracy because this package "owns" `Colorant`.

However, I think this violates the intent of `floattype`; the new type needs to be a replacement of `T` that uses floating-point arithmetic. For `RGB{N0f8}`, `RGB{Float32}` would be acceptable. Resolve this with `base_colorant_type`.

Make sure you check behavior on non-parametric types like `ARGB32`.

src/ColorTypes.jl
johnnychen94: I plan to remove the direct dependent on `ColorTypes.jl` or `FixedPointNumbers.jl` for `Images.jl`.

Perhaps without `import FixedPointNumbers: floattype`, just define a new function with the same name and export it? I.e.,

```julia
export floattype

floattype(::Type{T}) where T<:AbstractFloat = T
floattype(::Type{T}) where T<:FixedPoint = FixedPointNumbers.floattype(T)
```

If FixedPointNumbers doesn't intend to export this name, it would be good though.

src/traits.jl
johnnychen94: perhaps what we need is `floattype_colorant` instead of `floattype` 🤔

timholy: `floattype(eltype(c))` is shorter, clearer, and doesn't define a new name one has to go look up and read the docstring on.

src/ColorTypes.jl
timholy: > I plan to remove the direct dependent on ColorTypes.jl or FixedPointNumbers.jl for Images.jl

For example by re-exporting the types from ImageCore? There might be some merit in that; if so we could export `floattype` from FixedPointNumbers.

johnnychen94: Yes, since ImageCore --reexport--> Colors --Reexport--> ColorTypes.

I can open a PR in FixedPointNumbers if you think this is worth doing.

johnnychen94: An alternative, `floattype_colorant` can be

```julia
_floattype(::Type{T}) where T <: AbstractFloat = T
_floattype(::Type{T}) where T <: Bool = Float32
_floattype(::Type{T}) where T <: FixedPoint = FixedPointNumbers.floattype(T)
floattype_colorant(::Type{T}) where T <: Colorant = base_colorant_type(T){_floattype(T)}
```

timholy: The whole purpose of `floattype` is to return a value that safe to act as an accumulator without overflow; that's what this function is doing, so it should be called `floattype`.

johnnychen94:
If there aren't further comments, I think this PR is ready to be merged.

PR #124, opened 2019-08-01
extend float for Colorant (closes #123)

### LINKED ISSUE #123: support `float` for Colorant
With `floattype` introduced in #121, one can promote the image from `Gray{N0f8}` to `Gray{Float32}` easily with the following code:

```julia
floattype(eltype(img)).(img)
```

While a perhaps more natural way of doing this is to extend `float` for `Colorant` and use

```julia
float.(img) # promote Gray{N0f8} to Gray{Float32}
```

P.S. I opened this issue because I made a silly mistake of not `import Base: float`. PR #124 achieves this.

johnnychen94:
With `floattype` introduced in #121, one can promote the image from `Gray{N0f8}` to `Gray{Float32}` easily with the following code:

```julia
floattype(eltype(img)).(img)
```

While a perhaps more natural and easier way of doing this is to extend `float` for `Colorant` and use

```julia
float.(img) # promote Gray{N0f8} to Gray{Float32}
```

### DIFF
--- a/src/ColorTypes.jl
+++ b/src/ColorTypes.jl
@@ -6,7 +6,7 @@ using Base: @pure
 
 const Fractional = Union{AbstractFloat, FixedPoint}
 
-import Base: ==, isapprox, hash, convert, eltype, length, show, oneunit, zero, reinterpret, getindex
+import Base: ==, isapprox, hash, convert, eltype, length, show, oneunit, zero, reinterpret, getindex, float
 using Random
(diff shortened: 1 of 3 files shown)


johnnychen94:
The test fails because this requires some conversion utils from ImageCore. This PR should be on `ImageCore` as an alternative to float32, float64, etc.

Close it now and will reopen it there.

Issue #139, opened 2019-12-07
Adding abstract types for gamut definitions


kimikage:
To fix issue #125, we need to reaffirm that the range of color space and the color gamut are based on different concepts. So why not add types for the gamuts?

Of course, specific gamut definitions and their traits are outside the scope of `ColorTypes.jl`. In `ColorTypes.jl`, only abstract types should be defined:
```julia
abstract type AbstractGamut end

abstract type AbstractRGBGamut <: AbstractGamut end

abstract type Gamut_sRGB <: AbstractRGBGamut end
```
We will consider their usage in `Colors.jl` later.

kimikage:
I will add the gamut types to Colors.jl, including `AbstractGamut`.
Although the problem with `rand` (#125, #150) remains, let's consider it separately.

Issue #125, opened 2019-09-01
`gamutmax(LCHab)` disagrees with the Colors.jl usage


kimikage:
The max saturation of `LCHab` is defined as `1` in ColorTypes.jl.
https://github.com/JuliaGraphics/ColorTypes.jl/blob/0d5904e19b41227938274d325886dff221a8e8d7/src/operations.jl#L16-L17

However, in Colors.jl, the saturation comes from the `a` and `b` of `Lab` which are _mostly_ in [-100,100].
https://github.com/JuliaGraphics/Colors.jl/blob/657e265a4cccc9fc32584f7f3505611537bfc79d/src/conversions.jl#L575-L580

Therefore, for the sake of convenience,  `100` (or `128`) is better for the max saturation, though the actual gamut is not so simple. (See also: #84)

timholy:
Another option is to remove support for ones without a clearly defined max/min?

kimikage:
In the case of `LCHab`, the removal is probably OK, as no one has pointed this out.

However, few types have a clearly defined max/min, because the range of color space and the color gamut are based on different concepts.

BTW, I found this issue when I ran `rand(LCHab)`.
I think `gamutmax` and `gamutmin` are practical when and only when they are used with `rand()` as historically they are introduced for such a purpose.
Perhaps what we should remove might be the term “gamut” rather than the function.

kimikage:
As mentioned above, I think that "color gamut" and `rand` should be separated.
Then, `rand` should be more comprehensive. For example, there is a problem with `TransparentColor`:
```julia
julia> rand(RGBA)
RGBA{Float64}(0.4054196785489861,0.5812337734513682,0.4739237429262706,0.35592317884458446)

julia> rand(Lab)
Lab{Float64}(49.35707185830977,54.78932461097497,-80.6404236285697)

julia> rand(LabA)
ERROR: MethodError: no method matching gamutmax(::Type{LabA{Float64}})
```

Although it is necessary to discuss whether `ColorTypes.jl` should specify the ranges (not gamuts) or not, it may be the duty of `ColorTypes.jl` to resolve the range of `AlphaColor` and `ColorAlpha`, given the range of `color_type`.

kimikage:
BTW, `gamutmin`/`gamutmux` for YIQ are broken, too.
[...]

PR #140, opened 2019-12-07
[RFC/WIP] Add abstract types for gamut definitions (#139)

kimikage:
This adds abstract types for the gamut definitions (cf. #139, #125).

This also changes the `gamutmin`/`gamutmax` definitions. The new `gamutmin`/`gamutmax` mean the corners of the *minimum bounding box* of the (implicitly) specified gamut. Specifying the gamut (i.e. "sRGB") specifies the range *somewhat* formally.

**Edit:**
Of course, if we allow breaking changes, we also have the option of moving these to `Colors.jl`. I think `rand()` (with gamuts) is too complex for `ColorTypes.jl`, a *minimal* package.

### DIFF
--- a/src/gamut.jl
+++ b/src/gamut.jl
@@ -35,7 +35,7 @@ function gamutmin(colorspace::Type{T},
     (gamutmin(color_type(T), Gamut_sRGB)..., 0)
 end
 
-function gamutin(::Type{T}, gamut::AbstractGamut) where {T<:Colorant}
+function gamutmin(::Type{T}, gamut::AbstractGamut) where {T<:Colorant}
     gamutmin(T, typeof(gamut)) # fallback
 end
 
--- a/src/gamut.jl
+++ b/src/gamut.jl
@@ -1,0 +1,88 @@
+
+abstract type AbstractGamut end
+
+abstract type AbstractRGBGamut <: AbstractGamut end
+
+abstract type Gamut_sRGB <: AbstractRGBGamut end
+
+"""
+    gamutmax(colorspace, gamut=Gamut_sRGB)
+
+Returns the upper bound of `gamut` in `colorspace`. If the `gamut` exceeds the
+upper bound of the `colorspace`, the latter is returned.
+"""
+gamutmax(::Type{T}) where {T<:Color} = gamutmax(T, Gamut_sRGB)
+
+function gamutmax(colorspace::Type{T},
+                  gamut::Type{<:AbstractGamut}=Gamut_sRGB) where {T<:TransparentColor}
+    (gamutmax(color_type(T), Gamut_sRGB)..., 1)
+end
+
+function gamutmax(::Type{T}, gamut::AbstractGamut) where {T<:Colorant}
+    gamutmax(T, typeof(gamut)) # fallback
+end
+
+"""
+    gamutmin(colorspace, gamut=Gamut_sRGB)
+
+Returns the lower bound of `gamut` in `colorspace`. If the `gamut` exceeds (is
+below) the lower bound of the `colorspace`, the latter is returned.
+"""
+gamutmin(::Type{T}) where {T<:Color} = gamutmin(T, Gamut_sRGB)
+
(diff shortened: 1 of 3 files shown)


kimikage:
We can generate the random `TransparentColor`s. (cf. https://github.com/JuliaGraphics/ColorTypes.jl/issues/125#issuecomment-557884348)
```julia
julia> rand(LabA)
LabA{Float64}(22.480820110815625,13.641014129939421,-5.753612633740147,0.23049096545302006)
```

We can override the default range of color space (in a somewhat tricky manner).
```julia
julia> ColorTypes.gamutmax(::Type{<:Lab}) = (100,200,200)

julia> rand(Lab)
Lab{Float64}(76.07250064744058,192.4625550050896,5.393193026026793)
```

src/gamut.jl
```diff
@@ -0,0 +1,88 @@
...
+gamutmin(::Type{<:DIN99d}, ::Type{Gamut_sRGB}) = (0,-38.1436,-45.8498)
+
+gamutmax(::Type{<:DIN99o}, ::Type{Gamut_sRGB}) = (99.9997,45.5242,44.3662)
+gamutmin(::Type{<:DIN99o}, ::Type{Gamut_sRGB}) = (0,-40.1101,-40.4900)
+
+# LMS colorspace depends on color appearance models.
+gamutmax(::Type{<:LMS}, ::Type{Gamut_sRGB}) = (0.9493,1.0354,1.0872)
+
+gamutmax(::Type{<:YIQ}, ::Type{Gamut_sRGB}) = (1,0.5957,0.5226)
+gamutmin(::Type{<:YIQ}, ::Type{Gamut_sRGB}) = (0,-0.5957,-0.5226)
```
kimikage: The color primaries of NTSC and sRGB are different, and the NTSC color gamut is actually wider than sRGB. However, this difference is often ignored as "RGB" does not conform the sRGB.
Therefore, the following measures can be considered:
- Set the default color gamut for `YIQ` to `Gamut_NTSC`
- Narrow the range to be correct as sRGB

In any case, the gamut conflicts with the conversion methods in `Colors.jl`.

src/gamut.jl
```diff
@@ -0,0 +1,88 @@
...
+
+gamutmax(::Type{<:YCbCr}, ::Type{Gamut_sRGB}) = (235,240,240)
+gamutmin(::Type{<:YCbCr}, ::Type{Gamut_sRGB}) = (16,16,16)
+
+gamutmax(::Type{<:AbstractGray}, ::Type{Gamut_sRGB}) = (1,)
+gamutmin(::Type{<:AbstractGray}, ::Type{Gamut_sRGB}) = (0,)
```
kimikage: Strictly speaking, sRGB should not specify the *so-called* gray range. The range of luminance `Y` is [0,1], but `Y` is different from the *so-called* gray.

kimikage:
I'm going to rename "gamut.jl" -> "gamuts.jl".

johnnychen94:
I like the simplicity of this design 👍

src/gamut.jl
```diff
@@ -0,0 +1,88 @@
...
+below) the lower bound of the `colorspace`, the latter is returned.
+"""
+gamutmin(::Type{T}) where {T<:Color} = gamutmin(T, Gamut_sRGB)
+
+function gamutmin(colorspace::Type{T},
+                  gamut::Type{<:AbstractGamut}=Gamut_sRGB) where {T<:TransparentColor}
+    (gamutmin(color_type(T), Gamut_sRGB)..., 0)
+end
+
+function gamutin(::Type{T}, gamut::AbstractGamut) where {T<:Colorant}
```
johnnychen94:
```suggestion
function gamutmin(::Type{T}, gamut::AbstractGamut) where {T<:Colorant}
```

kimikage:
I prioritize extending Colors.jl and close this PR.(cf. https://github.com/JuliaGraphics/ColorTypes.jl/issues/139#issuecomment-593061604)

PR #143, opened 2019-12-17
Rename RGB1->XRGB, RGB4->RGBX

timholy:
I realized we should do this sooner than later, to give other packages time to deal with deprecation warnings. When we release 1.0 we should eliminate the deprecations.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ColorTypes"
 uuid = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
-version = "0.8.1"
+version = "0.9.0"
 
 [deps]
 FixedPointNumbers = "53c48c17-4a7d-5ca2-90c5-79b7896eea93"
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ ARGB(RGB(red, green, blue), alpha)
 
 ## Colors
 
-### RGB plus BGR, RGB1, RGB4, and RGB24: the AbstractRGB group
+### RGB plus BGR, XRGB, RGBX, and RGB24: the AbstractRGB group
 
 The [sRGB colorspace](https://en.wikipedia.org/wiki/SRGB).
 
@@ -92,7 +92,7 @@ crucial point: **for all `AbstractRGB` types, the constructor
 accepts values in the order `(r,g,b)` regardless of how they
 are arranged internally in memory**.
 
-`RGB1` and `RGB4` seem exactly like `RGB`, but internally they insert
+`XRGB` and `RGBX` seem exactly like `RGB`, but internally they insert
 one extra ("invisible") padding element; when the element type is
 `N0f8`, these have favorable memory alignment for interfacing with
 libraries like OpenGL.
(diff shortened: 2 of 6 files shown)


tknopp:
@timholy: Were there any deprecation warnings?

timholy:
Of course :smile:

tknopp:
Hm, was that ironic? People are running through my institute reporting that something broke and we were able tracking it down to hear. I then read your OP, which sounded that deprecation warnings actually should have been emitted.

timholy:
Definitely not ironic. You know you can check the `Files changed` tab yourself without having to ask---the depwarns are in the top few lines of `ColorTypes.jl`.

Some folks don't think of depwarns as breaking, but we bumped the minor version number anyway. If you're using upper bounds on package versions, as is recommended now, nothing should have changed until you were ready to upgrade.

More specifics about the breakage would be useful. The only thing that should break is performance (which will go into the toilet until you change the names).

tknopp:
> Definitely not ironic. You know you can check the Files changed tab yourself without having to ask---the depwarns are in the top few lines of ColorTypes.jl.

Sorry, I actually did exactly that, but for some reason did oversaw them, sorry! 

The upper bounds thing is clear, but we might have done something wrong in this package https://github.com/tknopp/ImageUtils.jl. Will have a look myself before I can give a better informed answer.

timholy:
There's another possible issue: if you get these via `using Colors` (rather than `using ColorTypes`) it looks like `@reexport` doesn't automatically export the deprecated names. Try upgrading to https://github.com/JuliaGraphics/Colors.jl/pull/383.

timholy:
It's bizarre, that package depends on Colors which shouldn't have allowed the upgrade without https://github.com/JuliaGraphics/Colors.jl/pull/383.

Oh, wait, I bet this is a registry bug, see https://github.com/JuliaRegistries/General/pull/6834

tknopp:
Here is the travis log that shows the error

https://travis-ci.org/tknopp/ImageUtils.jl/jobs/626828413

tknopp:
So was the bug that Colors had no upper bound and therefore claimed to be compatible with future ColorTypes? Because ImageUtlis should only have a version contract with Colors not ColorTypes.

tknopp:
This is actually a very interesting case that showcases an important scalability use case.

I think of Colors (or Images) as a metapackage that is responsible for keeping a consistent interface. These metapackge will be even more important in the "upper bound" scenario, since they are responsible for not breaking the interfaces. Thus, new changes first need to be tested in the metapackage, which makes only a release with larger upper bounds, one all of its tests pass.

tknopp:
There is still something wrong:

https://travis-ci.org/tknopp/ImageUtils.jl/jobs/626828416

The issue comes from ImageCore.

timholy:
https://github.com/JuliaRegistries/General/pull/6903

timholy:
That's merged now, you could try again.

tknopp:
I don't know if I need to do something specific but just restarting travis, the error persists:

https://travis-ci.org/tknopp/ImageUtils.jl/jobs/626828413

timholy:
Dang. https://github.com/JuliaRegistries/General/pull/6905

timholy:
That was very sneaky of the resolver to install ImageCore 0.8.2. Without an algorithm to set these retrospective bounds, I am not entirely confident of my ability to block all possible escape hatches.

tknopp:
you got it: https://travis-ci.org/tknopp/ImageUtils.jl

PR #146, opened 2020-01-09
[RFC] Move type aliases `AbstractAGray`/`AbstractGrayA` from Colors.jl

kimikage:
cf. https://github.com/JuliaGraphics/Colors.jl/issues/273#issuecomment-557839812

This also adds new type aliases `AbstractARGB`/`AbstractRGBA` for the consistency. (This does not add `AlphaColor3`/`ColorAlpha3`, though.)
`ARGB32` and `AGray32` are now defined with the new type aliases.

### DIFF
--- a/src/ColorTypes.jl
+++ b/src/ColorTypes.jl
@@ -16,6 +16,7 @@ export Fractional
 export Colorant
 export Color, TransparentColor, AlphaColor, ColorAlpha, AbstractRGB
 export AbstractGray, Color3, TransparentGray, Transparent3, TransparentRGB, ColorantNormed
+export AbstractAGray, AbstractGrayA, AbstractARGB, AbstractRGBA
 
 export RGB, BGR, XRGB, RGBX
 export HSV, HSB, HSL, HSI
--- a/src/types.jl
+++ b/src/types.jl
@@ -71,6 +71,10 @@ TransparentGray{C<:AbstractGray,T} = TransparentColor{C,T,2}
 Transparent3{C<:Color3,T}          = TransparentColor{C,T,4}
 TransparentRGB{C<:AbstractRGB,T}   = TransparentColor{C,T,4}
 ColorantNormed{T<:Normed,N}        = Colorant{T,N}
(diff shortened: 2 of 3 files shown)


kimikage:
BTW, these alias definitions do not have `const`, is there any reason for this? Is it simplicity?

kimikage:
Also, I cannot decide whether to add "the" workaround before merging this PR. :dizzy_face:
If we decide to add the workaround, we have to create a branch for v0.8.2. So, if anyway we make at least one branch, it is also a reasonable idea to make two branches *on the demand*.

timholy:
> BTW, these alias definitions do not have const, is there any reason for this? Is it simplicity? Or the remnant of @compat (for typealias keyword)?

I'm pretty sure that with type parameters on the LHS the `const` shouldn't be necessary---Julia knows what this is. But it doesn't hurt anything to add them, if you want. https://github.com/JuliaLang/julia/pull/20500

kimikage:
I personally prefer to always attach `const` to prevent forgetting to write `const`, but I'll follow the current short style.

kimikage:
I have been checking the segfault problem with JLD2 which occurs in my environment to confirm the need for segfault workaround for ColorTypes.jl. However I have not found a way to reproduce it.
The CI tests of JLD2 have been passed certainly. Paradoxically, this means that the green signal of CI does not guarantee the safety against the segfault.

I'm not blaming anyone, but to be honest, I'm quite sick of the problem.

kimikage:
With ColorTypes v0.9.1 released, including the workaround, we can make breaking changes without hesitation.

timholy:
You'd argue that this needs to be 0.10 due to the new name conflict with Colors.jl, right?

kimikage:
That's right. I also think we should merge PR #147 before releasing v0.10.
However, I was lost in the thick forest #149. :dizzy_face:
For now, we have the option of putting the additional tests into the forest (i.e. "runtests.jl") and quickly releasing v0.10.

PR #147, opened 2020-01-09
Move promotion rules from Colors.jl and Add gray-to-rgb conversion methods

### LINKED ISSUE #144: Adding/migration of gray-to-rgb conversion methods
ColorTypes.jl has the special real-to-rgb conversion methods.
https://github.com/JuliaGraphics/ColorTypes.jl/blob/81682158aec852502755ca0d3b943d2f34a71678/src/conversions.jl#L43-L44
In addition, ColorTypes.jl assumes that real numbers are gray-like and vice versa.
https://github.com/JuliaGraphics/ColorTypes.jl/blob/81682158aec852502755ca0d3b943d2f34a71678/src/conversions.jl#L49-L51
Therefore, ColorTypes.jl "substantially" defines the gray-to-rgb conversions. However, ColorTypes.jl does not actually provide the gray-to-rgb conversion methods.
```julia
julia> using ColorTypes

julia> convert(RGB24, 0.8)
RGB24(0.8N0f8,0.8N0f8,0.8N0f8)

julia> convert(RGB24, Gray24(0.8))
ERROR: No conversion of Gray24(0.8N0f8) to RGB24 has been defined

julia> convert(RGB, Gray(0.8))
ERROR: No conversion of Gray{Float64}(0.8) to RGB{Float64} has been defined

julia> using Colors

julia> convert(RGB, Gray(0.8))
RGB{Float64}(0.8,0.8,0.8)
```

I think the rgb-to-gray conversion methods should be in Colors.jl, but the gray-to-rgb conversion methods may be in ColorTypes.jl.
The reason I think so is that it is time to do https://github.com/JuliaGraphics/Colors.jl/issues/273#issuecomment-557839812. :rocket:

**Edit:**
Of course, leaving `promotion_rule`s in Colors.jl is another option. By using only ColorTypes.jl and not using Colors.jl, we can define different promotion rules from Colors.jl. However, in practice, it is not a good thing.

kimikage:
~This is a draft PR for issue #144. (This depends on PR #146.)~ PR #146 was merged.

This moves the promotion rules with grays and rgbs from Colors.jl.(Closes #144)
This also adds the gray-to-rgb conversion methods.

The current promotion rules have some problems, but this does not fix them since this focuses on the migration.

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -1,6 +1,13 @@
 Base.promote_rule(::Type{T1}, ::Type{T2}) where {T1<:AbstractGray,T2<:AbstractGray} = Gray{promote_type(eltype(T1), eltype(T2))}
 Base.promote_rule(::Type{T1}, ::Type{T2}) where {T1<:AbstractRGB,T2<:AbstractRGB} = RGB{promote_type(eltype(T1), eltype(T2))}
 
+Base.promote_rule(::Type{C3}, ::Type{Cgray}) where {C3<:Color3,Cgray<:AbstractGray} = base_colorant_type(C3){promote_type(eltype(C3), eltype(Cgray))}
+Base.promote_rule(::Type{C3}, ::Type{Cagray}) where {C3<:Color3,Cagray<:AbstractAGray} = alphacolor(base_colorant_type(C3)){promote_type(eltype(C3), eltype(Cagray))}
+Base.promote_rule(::Type{C3}, ::Type{Cgraya}) where {C3<:Color3,Cgraya<:AbstractGrayA} = coloralpha(base_colorant_type(C3)){promote_type(eltype(C3), eltype(Cgraya))}
+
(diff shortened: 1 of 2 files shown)


kimikage:
To merge this, I added some tests, and then I noticed the following issues:
```julia
julia> using Colors

julia> promote(RGB24(0.2,0.3,0.4), Gray(0.8))
ERROR: TypeError: in Type{...} expression, expected UnionAll, got Type{RGB24}

julia> promote(RGB24(0.2,0.3,0.4), GrayA(0.8))
ERROR: MethodError: no method matching coloralpha(::Type{RGB24})

julia> promote(RGB24(0.2,0.3,0.4), AGray(0.8))
ERROR: TypeError: in Type{...} expression, expected UnionAll, got Type{ARGB32}
```

However, since this PR focuses on the migration, I will not fix them.

kimikage:
Modifications to `color` may affect these additional tests. However, it is the good practice in TDD to add tests before writing or modifying the codes.
Therefore, I will merge this PR first.

Issue #157, opened 2020-02-04
Problem of constructing ABGR/BGRA with integer arguments


kimikage:
The following constructors generated by the macro are actually interpreted as `(b, g, r, a)`.
```julia
(::Type{ABGR})(r::T, g::T, b::T, alpha::T=1) where T<:Integer
(::Type{BGRA})(r::T, g::T, b::T, alpha::T=1) where T<:Integer
```
https://github.com/JuliaGraphics/ColorTypes.jl/blob/2c6bd3451988a974a1008c61a58dcd599a754553/src/types.jl#L499
https://github.com/JuliaGraphics/ColorTypes.jl/blob/2c6bd3451988a974a1008c61a58dcd599a754553/src/types.jl#L516

`($(fields...), alpha)` is expanded to `(b, g, r, a)`, and it is wrong as the constructor fields.

![bgra](https://user-images.githubusercontent.com/12679384/73729248-07c1a400-4778-11ea-924b-abe500e33c7f.png)

Issue #156, opened 2020-02-04
Constructing transparent colors with FixedPoint arguments


kimikage:
We can construct non-rgb colors with FixedPoint arguments (cf. issue #80, PR #81).
```julia
julia> HSV(1N0f8, 0.5N0f8, 0N0f8)
HSV{Float32}(1.0f0,0.5019608f0,0.0f0)
```

However, we cannot construct non-rgb transparent colors.
```julia
julia> AHSV(1N0f8, 0.5N0f8, 0N0f8)
ERROR: TypeError: in AHSV, in T, expected T<:AbstractFloat, got Type{Normed{UInt8,8}}

julia> HSVA(1N0f8, 0.5N0f8, 0N0f8, 0.2N0f8)
ERROR: TypeError: in HSVA, in T, expected T<:AbstractFloat, got Type{Normed{UInt8,8}}
```
Perhaps they are not needed, but that is inconsistent.

Issue #154, opened 2020-02-02
Minor problems with `color3types` and `parametric3`


kimikage:
I am *still* working on the issue #149.:sweat_smile:
ColorTypes has two collections of types: `color3types` and `parametric3`.
https://github.com/JuliaGraphics/ColorTypes.jl/blob/53c59257c3dddbebb05a899324a202a89d5ed213/src/types.jl#L417-L426
Lexically, `parametric3` is a subset of `color3types`. However, they are actually the same.
#### Julia v1.3.1 + ColorTypes v0.9.1
<details>
<summary>ColorTypes.color3types == ColorTypes.parametric3</summary>

```julia
julia> ColorTypes.color3types
20-element Array{UnionAll,1}:
 BGR
 DIN99
 DIN99d
 DIN99o
 HSV
 HSI
 HSL
 HSV
 LCHab
 LCHuv
 LMS
 Lab
 Luv
 RGB
 RGBX
 XRGB
 XYZ
 YCbCr
 YIQ
 xyY
```
</details>

#### Julia v0.5.2 + ColorTypes v0.5.2
<details>
<summary>ColorTypes.color3types == ColorTypes.parametric3</summary>

```julia
julia> ColorTypes.color3types
20-element Array{DataType,1}:
 ColorTypes.BGR{T<:Union{AbstractFloat,FixedPointNumbers.FixedPoint}}
 ColorTypes.DIN99{T<:AbstractFloat}
 ColorTypes.DIN99d{T<:AbstractFloat}
 ColorTypes.DIN99o{T<:AbstractFloat}
 ColorTypes.HSV{T<:AbstractFloat}
 ColorTypes.HSI{T<:AbstractFloat}
 ColorTypes.HSL{T<:AbstractFloat}
 ColorTypes.HSV{T<:AbstractFloat}
 ColorTypes.LCHab{T<:AbstractFloat}
 ColorTypes.LCHuv{T<:AbstractFloat}
 ColorTypes.LMS{T<:AbstractFloat}
 ColorTypes.Lab{T<:AbstractFloat}
 ColorTypes.Luv{T<:AbstractFloat}
 ColorTypes.RGB{T<:Union{AbstractFloat,FixedPointNumbers.FixedPoint}}
 ColorTypes.RGB1{T<:Union{AbstractFloat,FixedPointNumbers.FixedPoint}}
[...]

Issue #153, opened 2020-01-18
What should `base_color_type(TransparentGray)` return?


kimikage:
As with #152, there are "undocumented" differences between `base_color_type` and `color_type`.
```julia
julia> base_color_type(TransparentGray) # Umm...
Color

julia> base_color_type(TransparentRGB)
AbstractRGB

julia> color_type(TransparentGray) === AbstractGray # Color{T,1} where T
true

julia> color_type(TransparentRGB) === AbstractRGB
true
```

kimikage:
Of course, this is not only with `TransparentGray`.
```julia
julia> color_type(Color3) === Color3
true

julia> base_color_type(Color3)
Color

julia> color_type(Transparent3) === Color3
true

julia> base_color_type(Color3)
Color
```

PR #158, opened 2020-02-08
Organize test cases (Fixes #149)

### LINKED ISSUE #149: Organizing test cases
Although it is what probably many people, including me, are pretending not to see, we should organize the test cases.

I'm going to add dozens of tests to merge PR #147. I can't stand the number of lines in a single file any more. In addition, the tests which I'm going to add, should have been included in either of tests for ColorTypes.jl or Colors.jl . However, I did not notice the absence until now. It is difficult to confirm the completeness of the current ColorTypes.jl's tests.

I want to split the tests into multiple files, and add `@testset`. In order to prevent the loss due to the move, I'll make commits in small units. This is a bit time-consuming, i.e. I can't merge PR #147 right away. If anyone do other work in ColorTypes.jl or Colors.jl, be aware of the consistency between them.

kimikage:
This PR splits the tests into multiple files and adds `@testset`s and some extra tests. (Fixes #149)

I noticed some issues (cf. #151, #152, #153, #154, #156, #157), but this PR does not make any changes in "src".

To make it easier to track test cases, I created commits in small increments. Once confirmed, you can squash the commits.

This PR still leaves the inconsistencies in the test cases and notations. However, this PR may help a little to know the inconsistencies and understand where we need to fix.

### DIFF
--- a/test/traits.jl
+++ b/test/traits.jl
@@ -484,6 +484,79 @@ end
     @test_throws ErrorException reinterpret(ARGB{N0f8}, 0x12345678)
 end
 
+@testset "comparisons" begin
+    Cp3 = ColorTypes.parametric3
+    for C in unique(vcat(Cp3, coloralpha.(Cp3), alphacolor.(Cp3)))
+        @test C{Float64}(1,0,0) == C{Float32}(1,0,0)
+        @test C{Float32}(1,0,0) != C{Float32}(1,0,0.1)
+    end
+
+    for (a, b) in ((Gray(1.0), Gray(1)),
+                   (GrayA(0.8, 0.6), AGray(0.8, 0.6)),
+                   (RGB(1, 0.5, 0), BGR(1, 0.5, 0)),
+                   (RGBA(1, 0.5, 0, 0.8), ABGR(1, 0.5, 0, 0.8)),
+                   (Gray(0.8N0f8), Gray24(0.8)))
+        local a, b
+        @test a !== b
+        @test a == b
+        @test hash(a) == hash(b)
+    end
+    for (a, b) in ((RGB(1, 0.5, 0), RGBA(1, 0.5, 0, 0.9)),
+                   (RGB(0.5, 0.5, 0.5), Gray(0.5)),
+                   (HSV(0, 0, 0.5), HSV(100, 0, 0.5)), # grays
+                   (Lab(70, 0, 60), LCHab(70, 60, 90)))
+        local a, b
+        @test a != b
+        @test hash(a) != hash(b)
+    end
+    # It's not obvious whether we want these to compare as equal, but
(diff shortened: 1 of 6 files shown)


test/traits.jl
```diff
@@ -0,0 +1,620 @@
...
+    @test reinterpret(Gray24, 0xdc00ff12) ==  Gray24(0.071)
+    @test reinterpret(UInt32, reinterpret(Gray24, 0xdc00ff12)) === 0xdc00ff12
+
+    @test reinterpret(AGray32, 0xab121212) === AGray32(0.071, 0.671)
+
+    @test_throws MethodError reinterpret(UInt32, ARGB{N0f8}())
+    @test_throws ErrorException reinterpret(ARGB{N0f8}, 0x12345678)
+end
+
+@testset "comparisons" begin
```
timholy: From here I'd say this is less `traits` and more `operations`

kimikage: I agree with you. However I categorized the tests based on the locations of target functions.

The `==` and `≈` are definitely "operators", but, I think, `isapprox` is something like "traits". Although both definitions should be close, I think they should be placed before "src/operations.jl" because of their primitivity.
https://github.com/JuliaGraphics/ColorTypes.jl/blob/2c6bd3451988a974a1008c61a58dcd599a754553/src/ColorTypes.jl#L43-L47
So, what about moving them later to a new file "comparison.jl"?

kimikage: A similar problem exists with `alphacolor`/`coloralpha`. However, since both "for types" are needed to define the constructors, I think they should still be in "src/types.jl".

kimikage:
To merge the improvement of `ccolor` and PR #147 early, I'll merge this PR first.

test/traits.jl
```diff
@@ -0,0 +1,620 @@
...
+    @test HSVA(100,0.4,0.6,0.8) ≈ HSVA{Float32}(100,0.4,0.6,0.8)
+end
+
+@testset "identities for Gray" begin
+    @test oneunit(Gray{N0f8}) === Gray{N0f8}(1)
+    @test zero(Gray{N0f8}) === Gray{N0f8}(0)
+    @test oneunit(Gray) === Gray{N0f8}(1)
+    @test zero(Gray) === Gray{N0f8}(0)
+
+    @test_throws MethodError oneunit(Gray24)
```
timholy, line 603: Out of curiosity, why shouldn't this be defined?

kimikage: I thought the same question, and I didn't understand the reason, so I followed the implementation.:sweat_smile:

timholy: It was probably an oversight. Better might have been `@test_broken oneunit(Gray24) === Gray24(1)`. I'll  add the missing methods to #186, since it also moves these methods.

Issue #151, opened 2020-01-17
Problem with `eltype(RGB)` or showing strange "RGB{Any}"


kimikage:
See: https://github.com/JuliaGraphics/Colors.jl/issues/369

I guess that the direct solution for JuliaGraphics/Colors.jl#369 is enhancing `zeros` in ColorVectorSpace.jl. (Of course, *only for `Gray`*, doing that in ColorTypes.jl is another option.)

Well, in the first place, `Gray{Any}` and `RGB{Any}` look strange to me.
```julia
julia> using ColorTypes

julia> eltype(RGB)
Any

julia> eltype(AGray)
Any

julia> typeof(RGB)
UnionAll

julia> typeof(RGB{Float32})
DataType
```
Yes, this is a known issue.
https://github.com/JuliaGraphics/ColorTypes.jl/blob/53c59257c3dddbebb05a899324a202a89d5ed213/test/runtests.jl#L27-L28

Of course, I don't really want to get `"Gray{<:Union{Bool, AbstractFloat, FixedPointNumbers.FixedPoint}}"` in the type summary. :smile:

Should we fix anything? Or will we continue to use `"Any"` for convenience?

kimikage:
If we regard this as a display problem, perhaps we need to fix `colorant_string_with_eltype`.
https://github.com/JuliaGraphics/ColorTypes.jl/blob/53c59257c3dddbebb05a899324a202a89d5ed213/src/traits.jl#L219-L234

BTW, I think that `colorant_string`, `colorant_string_with_eltype` and `showcoloranttype` should be in "show.jl " rather than "traits.jl", because they are not exported. `showcoloranttype` should definitely be moved to "show.jl".

kimikage:
I don't like this recursive trick very much. Perhaps it was a necessary technique in older versions, though.
https://github.com/JuliaGraphics/ColorTypes.jl/blob/53c59257c3dddbebb05a899324a202a89d5ed213/src/traits.jl#L111-L116

However, it will take some care to handle `UnionAll`s properly. For example:
https://github.com/JuliaGraphics/ColorTypes.jl/blob/53c59257c3dddbebb05a899324a202a89d5ed213/src/traits.jl#L148-L154

timholy:
[...]

Issue #137, opened 2019-12-07
support `ones(::Type{<:AbstractGray}, sz)`


johnnychen94:
Since `one(Gray)` is deprecated, it would be convenient that the following kind of usage isn't affected:

```julia
julia> ones(Gray{N0f8}, 4, 4)
┌ Warning: one(Gray{Normed{UInt8,8}}) will soon switch to returning 1; you might need to switch to `oneunit`
│   caller = ones at array.jl:462 [inlined]
└ @ Core .\array.jl:462
4×4 Array{Gray{N0f8},2} with eltype Gray{Normed{UInt8,8}}:
...
```

It is "easier" to read than `Gray.(ones(N0f8, 4, 4))`.

P.S., this has the same issue in https://github.com/JuliaGraphics/Colors.jl/issues/369 if we don't specify the concrete storage type.

```julia
julia> ones(Gray, 4, 4)
4×4 Array{Gray{Any},2} with eltype Gray:
```

kimikage:
cf. #130

johnnychen94:
Perhaps it's better to move this issue to `ColorVectorSpace`

johnnychen94:
cc @ianshmean

Just saw your post in slack

kimikage:
Should we include a solution for this issue in ColorType.js v0.10? (cf. https://github.com/JuliaGraphics/ColorTypes.jl/pull/160#issuecomment-592506240)

cc @johnnychen94

johnnychen94:
It looks like ColorTypes is the right place. I'll see if I can manage it tomorrow if you don't have plans on covering it. (My recent focus has shifted to Flux and deep learning things 😄 )

kimikage:
This was supported by PR #243. (I thought I had closed this.:sweat_smile:)
```julia
julia> ones(Gray{N0f8}, 4, 4)
4×4 Matrix{Gray{N0f8}}:
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0

julia> ones(Gray, 4, 4) # non-concrete type
4×4 Matrix{Gray}:
 Gray{N0f8}(1.0)  Gray{N0f8}(1.0)  Gray{N0f8}(1.0)  Gray{N0f8}(1.0)
[...]

PR #160, opened 2020-02-09
add isless(a::AbstractGray, b::AbstractGray)

alyst:
It's required by JuliaImages/ImageMorphology.jl#24 to support MaxTree-based operations in grayscale images.

### DIFF
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -385,3 +385,6 @@ function Base.one(::Type{C}) where {C<:Gray}
 end
 
 Base.broadcastable(x::Colorant) = Ref(x)
+
+Base.isless(a::AbstractGray, b::AbstractGray) =
+    isless(gray(a), gray(b))
--- a/test/traits.jl
+++ b/test/traits.jl
@@ -594,6 +594,49 @@ end
     @test HSVA(100,0.4,0.6,0.8) ≈ HSVA{Float32}(100,0.4,0.6,0.8)
 end
 
+@testset "isless" begin
+    @test Gray(0.8) < Gray(0.9)
+    @test Gray(0.8) <= Gray(0.9)
+    @test Gray(0.9) <= Gray(0.9)
+    @test Gray(0.9) > Gray(0.8)
+    @test Gray(0.9) >= Gray(0.8)
+    @test Gray(0.9) >= Gray(0.9)
+
+    @test Gray(0.8f0) < Gray(0.9f0)
+    @test Gray(0.8f0) <= Gray(0.9f0)
+    @test Gray(0.9f0) <= Gray(0.9f0)
+    @test Gray(0.9f0) > Gray(0.8f0)
+    @test Gray(0.9f0) >= Gray(0.8f0)
+    @test Gray(0.9f0) >= Gray(0.9f0)
+
+    @test Gray(0.8N0f8) < Gray(0.9N0f8)
+    @test Gray(0.8N0f8) <= Gray(0.9N0f8)
+    @test Gray(0.9N0f8) <= Gray(0.9N0f8)
+    @test Gray(0.9N0f8) > Gray(0.8N0f8)
+    @test Gray(0.9N0f8) >= Gray(0.8N0f8)
+    @test Gray(0.9N0f8) >= Gray(0.9N0f8)
+
+    @test Gray(0.8) < Gray(0.9f0)
+    @test Gray(0.8) <= Gray(0.9f0)
+    # @test Gray(0.9) <= Gray(0.9f0) is not true due to approximation
+    @test Gray(0.9) > Gray(0.8f0)
+    @test Gray(0.9) >= Gray(0.8f0)
+    @test Gray(0.9) >= Gray(0.9f0)
+
+    @test Gray(0.8f0) < Gray(0.9N0f8)
+    @test Gray(0.8f0) <= Gray(0.9N0f8)
+    @test Gray(0.9f0) <= Gray(0.9N0f8)
+    @test Gray(0.9f0) > Gray(0.8N0f8)
+    @test Gray(0.9f0) >= Gray(0.8N0f8)
+    # @test Gray(0.9f0) >= Gray(0.9N0f8) is not true, since 0.9N0f8 = 0.902
+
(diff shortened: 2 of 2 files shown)


johnnychen94:
This is already covered by another package `ColorVectorSpace`

timholy:
It occurs to me that perhaps we should have ImageCore load ColorVectorSpace, just so it's always something we can count on. Thoughts?

alyst:
> This is already covered by another package ColorVectorSpace.jl

What about moving `isless(Gray, Gray)` here? There's already `hash`, `isapprox`, '=='.
It's a bit confusing that the type is defined here, but the basic one-line functions are implemented elsewhere.
I can understand why RGB support requires dedicated package, but `Gray` is very straightforward.

timholy:
Yeah, maybe we should have it here, given the other comparison functions. I'll leave this open for a bit to see whether objections are raised.

kimikage:
> (To be honest, I'm not comfortable with treating Gray like a dimensionless number.)

_Originally posted by @kimikage in https://github.com/JuliaGraphics/ColorTypes.jl/issues/144#issuecomment-574679534_

I have respect for the historical background and what we are actually using. Therefore, I have no objection to moving `isless`.
However, if we do it, I want to create a new file (cf. https://github.com/JuliaGraphics/ColorTypes.jl/pull/158#discussion_r376708399). :smile:

johnnychen94:
Loading `ColorVectorSpace` in `ImageCore` is the simplest way; moving `isless` here requires some non-trivial compatibility check. -- Note that downstream packages such as `ImageMorphology` doesn't specify compat version for `ColorTypes`.

kimikage:
You are right, @johnnychen94.
However, the next ColorTypes v0.10 is going to require the new Colors v0.12, which is incompatible with ColorTypes v0.9. This means that ColorVectorSpace should be bumped up, too. Thus, the chain of changes will be terminated there.

alyst:
I've took a look at what *ColorVectorSpace.jl* currently provides. To me it looks like ColVecSpace.jl could be merged into ColorTypes.jl, but I guess the major reason to split was to reduce compilation time.
Plus, ColVecSpace.jl introduces deps on packages like SpecialFunctions.jl that we don't want to introduce into ColorTypes.jl.
But maybe the alternative could be some hypothetic *Lift.jl* package that provides `@lift` macro, which automatically defines `f(a::Gray) = f(a.val)` for all `f` available in the supported packages of the active environment (using Requires.jl for optional dependencies on packages like SpecialFunctions.jl).

timholy:
> To me it looks like ColVecSpace.jl could be merged into ColorTypes.jl, but I guess the major reason to split was to reduce compilation time.

Not really. Its purpose is to give everyone what they want:

- people who want high-performance RGB processing load ColorVectorSpace
- people who want to be colorimetrically correct strenuously avoid loading ColorVectorSpace

The motivation is [here](https://github.com/JuliaGraphics/ColorVectorSpace.jl#introduction).

There would certainly be nothing wrong, then, with moving this particular bit of functionality here.

alyst:
> people who want to be colorimetrically correct strenuously avoid loading ColorVectorSpace

Would it make sense to implement both "fast" and "correct" transformations within the same package (*ColorTypes.jl*, eventually)?
Something like `sin(color1) + tanh(color2)` is fast, and `@colorimetric(sin(color1) + tanh(color2))` is colorimetrically correct?
My concern with the current design is that the active environment is used to define the behaviour of the calculations, whereas I though that the primary use of the environment is to track version dependencies.
What if *ColorVectorSpace.jl* is loaded (e.g. because we need it for grayscale image processing or for fast RGB processing), but the same module also needs colorimetrically correct transformations?

johnnychen94:
Speaking of loading ColorVectorSpaces, it would be useful if there's a flag to check whether it's in the fast or the correct mode, since many downstream packages would inexplicably load it.

alyst:
> Speaking of loading ColorVectorSpaces, it would be useful if there's a flag to check whether it's in the fast or the correct mode

IIUC the flag (global option) would mean that Julia would not be able to efficiently compile many function calls.
Plus, global flag would affect all uses of ColVecSpaces, including those that rely on their own settings.

kimikage:
In the first place, the arithmetic of colors cannot be generally defined. The "colorimetrically correct" way is not to use `sin(color1)`, `tanh(color2)` and so on.
For example, we can consider the arithmetic in XYZ space as the "colorimetrically correct" arithmetic. However, it may bother users who use another "correct" definition. So, it is better to create a new package `XYZColorVectorSpace.jl`, and it makes the situation worse.

kimikage:
When PR #171 is merged, we can remove the promotion rules, some typealiases and the methods for gray-->rgb conversions from Colors.jl (https://github.com/JuliaGraphics/Colors.jl/pull/403).
Sorry, I spent much time rewriting the tests and the promotion rules, so the FixedPointNumbers update on JuliaGraphics/JuliaImages packages has been delayed.

Therefore it's time to decide whether to do https://github.com/JuliaGraphics/ColorTypes.jl/pull/160#issuecomment-583868517.

timholy:
I favor merging this. Any objections?

kimikage:
https://github.com/JuliaGraphics/ColorTypes.jl/pull/160#issuecomment-583871264

> However, if we do it, I want to create a new file (cf. [#158 (comment)](https://github.com/JuliaGraphics/ColorTypes.jl/pull/158#discussion_r376708399)). 😄

johnnychen94:
The only problem I'm concerning is that we can't move all gray arithmetic here; https://github.com/JuliaGraphics/ColorVectorSpace.jl/issues/38 and sometimes later there will be another call on reexporting ColorVectorSpace...

timholy:
We don't need arithmetic for `isless`, though. Just a total order, which Gray values have.

timholy:
Creating a new file would be fine, but probably best after merging this?

kimikage:
When is this going to be decided? It may be better to submit the issue in ColorVectorSpace.

Also, what else should be included in ColorTypes v0.10? When we decide to merge this, what should we do about #137?
I personally want to change the "display" problem in #151, but there is no rush.

Issue #152, opened 2020-01-18
`GrayA{Bool}` and `AGray{Bool}` are unavailable


kimikage:
I don't know whether `GrayA{Bool}` and `AGray{Bool}` are really needed. I don't mind if they are invalid types, but there are some inconsistencies.
```julia
julia> using ColorTypes

julia> Gray{Bool}(1) # this is ok
Gray{Bool}(true)

julia> GrayA{Bool}(1,1)
ERROR: TypeError: in GrayA, in T, expected T<:Union{AbstractFloat, FixedPointNumbers.FixedPoint}, got Type{Bool}

julia> coloralpha(Gray{Bool}) # this is ok, but disquieting
GrayA

julia> coloralpha(Gray{Bool}(1)) # gotcha!
ERROR: TypeError: in GrayA, in T, expected T<:Union{AbstractFloat, FixedPointNumbers.FixedPoint}, got Type{Bool}

julia> color_type(Gray)
Gray

julia> color_type(GrayA) # this is not wrong, but counterintuitive
Gray{T} where T<:Union{AbstractFloat, FixedPointNumbers.FixedPoint}

julia> color_type(GrayA) === Gray
false

julia> color_type(RGBA) === RGB # this is simple and intuitive
true
```

I don't fully understand why `Bool` is allowed only for `Gray` and not for `RGB`. I also think `Gray{Bool}` is useful for binary (black/white) images (cf. PR #35, issue #76, #29), but that is not a reason to prohibit `GrayA{Bool}` and `AGray{Bool}`.

kimikage:
What is this again?
https://github.com/JuliaGraphics/ColorTypes.jl/blob/53c59257c3dddbebb05a899324a202a89d5ed213/src/traits.jl#L176

kimikage:
If we relax the component types to `Real` in v0.13, this issue will be solved automatically. Of course, there are several problems with the relaxing, though.

PR #161, opened 2020-02-12
Improve implementation of ccolor

### LINKED ISSUE #153: What should `base_color_type(TransparentGray)` return?

### LINKED ISSUE #154: Minor problems with `color3types` and `parametric3`

timholy:
This does some maintenance on the type system of ColorTypes. In particular it leverages the huge improvements in Julia's type system and optimizer that allow one to write the code in block form rather than large dispatch chains. This makes it somewhat easier to perform complex logic. It fixes quite a few of our `@test_broken`s, though there are others that should be investigated.

With respect to upstream tests, the most breaking part of this might be the introduction of `TwoColorTypeError`; I can revert that and go back to `ErrorException` if there's a clear consensus against it. But I introduced it to improve printing of error messages for the user.

I have some additional tests I wrote, but they are partly redundant with what we already have. Therefore I'm posting them in a [gist](https://gist.github.com/timholy/21358b924b3a885e9d2a84f3eb3eb15e); folks can tell me whether we want them or not.

If https://github.com/JuliaLang/julia/issues/8322 gets implemented we should make further changes to how the types themselves are declared.

Fixes #153 
Fixes #154

### DIFF
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -197,52 +197,43 @@ base_colorant_type(::Type{C}) where {C<:Colorant} = isabstracttype(C) ? abstract
 abstract_basetype(::Type{C}) where C  <: AbstractRGB       = AbstractRGB
 abstract_basetype(::Type{C}) where C  <: ColorN{N} where N = ColorN{N}
 abstract_basetype(::Type{Color})                           = Color
-abstract_basetype(::Type{C}) where C                       = abstract_transparent_basetype(C)  # separate function to control method ordering
-
-function abstract_transparent_basetype(::Type{AC}) where AC
-    AC<:AlphaColor && return alphacolor_basetype(AC)
-    AC<:ColorAlpha && return coloralpha_basetype(AC)
-    AC<:TransparentColor && return transparentcolor_basetype(AC)
-    return colorant_basetype(AC)
-end
-
-alphacolor_basetype(::Type{AC}) where AC <: AlphaColorN{N,C} where {N,C<:Color}  = AlphaColor{base_colorant_type(C){T},T,N} where T
-coloralpha_basetype(::Type{AC}) where AC <: ColorAlphaN{N,C} where {N,C<:Color}  = ColorAlpha{base_colorant_type(C){T},T,N} where T
-transparentcolor_basetype(::Type{AC}) where AC <: TransparentColorN{N,C} where {N,C<:Color} =
+abstract_basetype(::Type{AC}) where AC <: AlphaColorN{N,C} where {N,C<:Color} = AlphaColor{base_colorant_type(C){T},T,N} where T
+abstract_basetype(::Type{CA}) where CA <: ColorAlphaN{N,C} where {N,C<:Color} = ColorAlpha{base_colorant_type(C){T},T,N} where T
+abstract_basetype(::Type{TC}) where TC <: TransparentColorN{N,C} where {N,C<:Color} =
     TransparentColor{base_colorant_type(C){T},T,N} where T
-transparentcolor_basetype(::Type{AC}) where AC <: TransparentColor               = TransparentColor
-colorant_basetype(::Type{C}) where C <: ColorantN{N} where N                     = ColorantN{N}
-colorant_basetype(::Type{C}) where C <: Colorant                                 = Colorant
-
+abstract_basetype(::Type{TC}) where TC <: TransparentColor = TransparentColor
 # These handle things like base_colorant_type(AbstractRGBA)
-function alphacolor_basetype(::Type{AC}) where AC <: AlphaColorN{N} where {N}
-    Cb = Base.parameter_upper_bound(AC, 1)
+parameter1(::Type{C}) where C = C isa DataType ? C.parameters[1] : Base.parameter_upper_bound(C, 1)
+function abstract_basetype(::Type{AC}) where AC <: AlphaColorN{N} where {N}
+    Cb = parameter1(AC)
     Cb === Color && return AlphaColor{C,T,N} where {T,C<:ColorN{N-1,T}}
+    isabstracttype(Cb) || return AlphaColor{Cb{T},T,N} where T
     return AlphaColor{C,T,N} where {T,C<:Cb{T}}
 end
-function coloralpha_basetype(::Type{AC}) where AC <: ColorAlphaN{N} where {N}
-    Cb = Base.parameter_upper_bound(AC, 1)
+function abstract_basetype(::Type{CA}) where CA <: ColorAlphaN{N} where {N}
+    Cb = parameter1(CA)
     Cb === Color && return ColorAlpha{C,T,N} where {T,C<:ColorN{N-1,T}}
+    isabstracttype(Cb) || return ColorAlpha{Cb{T},T,N} where T
     return ColorAlpha{C,T,N} where {T,C<:Cb{T}}
 end
-function transparentcolor_basetype(::Type{AC}) where AC <: TransparentColorN{N} where {N}
-    Cb = Base.parameter_upper_bound(AC, 1)
+function abstract_basetype(::Type{TC}) where TC <: TransparentColorN{N} where {N}
+    Cb = parameter1(TC)
     Cb === Color && return TransparentColor{C,T,N} where {T,C<:ColorN{N-1,T}}
+    isabstracttype(Cb) || return TransparentColor{Cb{T},T,N} where T
     return TransparentColor{C,T,N} where {T,C<:Cb{T}}
 end
+# This fallback dispatches to a separate function to control method ordering.
+# Otherwise the generic ColorantN methods supersede the AlphaColor/ColorAlpha/Transparent methods
+abstract_basetype(::Type{C}) where C = abstract_colorant_basetype(C)
 
-# # Ensure parametric return type
-# base_colorant_type(::Type{RGB24})   = RGB
-# base_colorant_type(::Type{Gray24})  = Gray
-# base_colorant_type(::Type{ARGB32})  = ARGB
-# base_colorant_type(::Type{AGray32}) = AGray
+abstract_colorant_basetype(::Type{C}) where C <: ColorantN{N} where N                     = ColorantN{N}
+abstract_colorant_basetype(::Type{C}) where C <: Colorant                                 = Colorant
 
 """
     Cbase = base_colorant_type(C::Type)
     Cbase = base_colorant_type(c::Colorant)
 
 Return the Colorant type without specified numeric element type.
-It always returns a parametric type, which makes it convenient for switching the numeric element type.
 
 For example, compare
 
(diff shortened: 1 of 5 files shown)


src/traits.jl
```diff
@@ -237,85 +303,119 @@ showcoloranttype(io, ::Type{Union{}}) = show(io, Union{})
...
+        eltype(C) <: Tdest && return C   # RGB24 etc.
+        error("nonparametric type ", C, " has ambiguous destination ", Cdest)
+    end
+    isabstracttype(Tdest) || !issupported(C, Tdest) || return C{Tdest}
+    T = pureintersect(Tdest, Tsrc)
+    T === Union{} && throw(TwoColorTypeError(:ccolor, "empty intersection between", Tdest, Tsrc))
+    T === Any && return C
+    Tdef = eltype_default(C)
+    issupported(C, T) && return C{T}
+    T<:Integer && return C{Tdef}
```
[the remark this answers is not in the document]
kimikage: BTW, the following may be a different issue. I'll report a new issue.
```julia
julia> ccolor(ARGB, Int)
ARGB{Normed{UInt8,8}}

julia> ARGB(1)
ERROR: MethodError: no method matching alpha(::Int64)
```

src/types.jl
```diff
@@ -414,15 +421,22 @@ AGray32(g::AbstractGray, alpha = 1) = AGray32(gray(g), alpha)
...
-    isdefined(ColorTypes, s) || return false
-    t = getfield(ColorTypes, s)
-    isa(t, Type) && t <: Colorant && !isabstracttype(t) && length(fieldnames(t))>1
-  end)
+                        filter(names(ColorTypes, all=false)) do s
+                            isdefined(ColorTypes, s) || return false
+                            t = getfield(ColorTypes, s)
+                            isa(t, Type) && t <: Color && !isabstracttype(t) && length(t) == 3
+                        end
+                        )
```
kimikage: I think the following is simpler and it is better to use `unique`.
```julia
filter(map(s->getfield(ColorTypes, s), names(ColorTypes, all=false))) do s
    isa(s, Type{<:Color3}) && !isabstracttype(s)
end |> unique
```

src/traits.jl
```diff
@@ -109,83 +109,149 @@ to_top(::Type{Colorant{T,N}}) where {T,N} = Colorant{T,N}
...
+abstract_basetype(::Type{C}) where C  <: ColorN{N} where N = ColorN{N}
+abstract_basetype(::Type{Color})                           = Color
+abstract_basetype(::Type{C}) where C                       = abstract_transparent_basetype(C)  # separate function to control method ordering
+
+function abstract_transparent_basetype(::Type{AC}) where AC
+    AC<:AlphaColor && return alphacolor_basetype(AC)
+    AC<:ColorAlpha && return coloralpha_basetype(AC)
+    AC<:TransparentColor && return transparentcolor_basetype(AC)
+    return colorant_basetype(AC)
+end
```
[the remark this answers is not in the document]
kimikage: In addition, is this (`abstract_transparent_basetype`) needed?
> separate function to control method ordering

What is the difference between the controlled order and the default dispatching order?

src/traits.jl
```diff
@@ -109,83 +109,149 @@ to_top(::Type{Colorant{T,N}}) where {T,N} = Colorant{T,N}
...
+eltypes_supported(::Type{RGB24})   = N0f8
+eltypes_supported(::Type{Gray24})  = N0f8
+eltypes_supported(::Type{ARGB32})  = N0f8
+eltypes_supported(::Type{AGray32}) = N0f8
 
 eltypes_supported(c::Colorant) = eltypes_supported(typeof(c))
 
-@pure issupported(::Type{C}, ::Type{T}) where {C<:Colorant,T} = T <: eltypes_supported(C)
+"""
+    issuported(C::Type, T::Type)::Bool
```
kimikage:
```suggestion
    issupported(C::Type, T::Type)::Bool
```

kimikage: It might be good idea to aggregate these eltype related functions around [here](https://github.com/JuliaGraphics/ColorTypes.jl/blob/5b2e50d62b8fd47a860c2e0543446144d8627c87/src/types.jl#L535-L549).

`eltype_ub` and `eltypes_supported` are different. I am concerned about using the reflection-like technique, despite the fact that we definitely know the supported eltypes. That also concerns #152.

src/types.jl
```diff
@@ -56,13 +56,13 @@ abstract type TransparentColor{C<:Color,T,N} <: Colorant{T,N} end
 alpha channel comes first in the internal storage order. **Note** that
 the constructor order is still `(color, alpha)`.
 """
-abstract type AlphaColor{C,T,N} <: TransparentColor{C,T,N} end
+abstract type AlphaColor{C<:Color,T,N} <: TransparentColor{C,T,N} end
```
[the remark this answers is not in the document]
timholy: More https://github.com/JuliaLang/julia/issues/6383. Specifying this helps dispatch tables. I made this change after working on the `abstract_basetype` hierarchy, and indeed this change makes some of the "dispatch to a different function" cascade no longer necessary. (Still need one case.)

src/types.jl
timholy: I took some of this, good suggestion.

But:
```julia
julia> module MissingDef
       export foo
       end
Main.MissingDef

julia> names(MissingDef)
2-element Array{Symbol,1}:
 :MissingDef
 :foo       

julia> getfield(MissingDef, :foo)
ERROR: UndefVarError: foo not defined
Stacktrace:
 [1] top-level scope at REPL[25]:1
```

test/traits.jl
```diff
@@ -418,10 +414,10 @@ end
     @test @inferred(ccolor(Colorant{N0f8}, RGBX{Float32})) === RGBX{N0f8}
     @test @inferred(ccolor(Colorant{N0f8,3}, BGR{N0f8})) === BGR{N0f8}
 
-    @test @inferred(ccolor(AbstractRGB, HSV{Float32})) === AbstractRGB{Float32} # is not concrete type
-    @test @inferred(ccolor(AbstractRGB{N0f8}, HSV{Float32})) === AbstractRGB{N0f8} # is not concrete type
+    @test_throws TwoColorTypeError ccolor(AbstractRGB, HSV{Float32})
+    @test_throws TwoColorTypeError ccolor(AbstractRGB{N0f8}, HSV{Float32})
     @test @inferred(ccolor(AbstractRGB, RGB24)) === RGB24
-    @test_broken @inferred(ccolor(AbstractRGB{Float32}, RGB24))
+    @test_throws ErrorException ccolor(AbstractRGB{Float32}, RGB24)
```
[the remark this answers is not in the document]
timholy: It's an interesting question the degree to which `ccolor` should mimic `promote_type`. Note that `promote_type` never throws an error, e.g.,
```julia
julia> promote_type(Complex{Float32}, String)
Any
```
Since `Any` is not concrete, `ccolor` cannot mimic `promote_type` exactly.

I've added some additional documentation about this.

timholy:
I believe I've address the comments, many thanks for an excellent review!

src/types.jl
kimikage: In general, I think it's good that errors occur. We can notice the problems and improve it. :smile:
Is there a reason that the definition should be missing in ColorTypes.jl?

timholy: I guess I always think of these as a pair, probably because I often write code on the compiler side of Julia where it's quite important to avoid making assumptions that all packages are perfect. But sure, as long as there's a suitable comment we could afford to be fragile here.

timholy: Oh, actually, we need my version after all: any exported name in this package that hasn't yet been loaded will cause an error:

```julia
julia> include("runtests.jl")
[ Info: Precompiling ColorTypes [3da002f7-5984-5a60-b8a6-cbb66c0b333f]
ERROR: LoadError: LoadError: UndefVarError: alpha not defined
Stacktrace:
 [1] #2 at /home/tim/.julia/dev/ColorTypes/src/types.jl:430 [inlined]
 [2] iterate at ./generator.jl:47 [inlined]
...
```
`alpha` isn't defined until `trails.jl` gets loaded.

timholy:
Any thoughts on whether we want any of https://gist.github.com/timholy/21358b924b3a885e9d2a84f3eb3eb15e? I haven't gone through to see if it improves test coverage, but I could try that if you want.

src/types.jl
kimikage: I have the opposite conclusion. This means that it depends on the order of loading, so your version has the "potential" risk of the absence of types. As you say, we have to think that as a pair.

kimikage:
I don't want to clean up test cases anymore:smile:, but I welcome adding "clean" test cases.

src/types.jl
timholy: I guess it's not as big a deal either way now. Older versions of this package defined some of the traits in terms of `color3types`, so you had to define its value early (before some traits were defined).

So it would be fine to change this if you want to move it to the end of the module definition.

PR #164, opened 2020-02-17
Improve promotion rules (Fixes #129)

### LINKED ISSUE #129: Promotion rules for same base color pairs
As https://github.com/JuliaGraphics/Colors.jl/issues/273 shows, there are the problems with promotion rules for same base color pairs. Here is another simple example:
```julia
julia> promote_type(HSV{Float64}, HSV{Float32})
HSV

julia> promote(HSV{Float64}(0,0,0), HSV{Float32}(0,0,0))
ERROR: promotion of types HSV{Float64} and HSV{Float32} failed to change any arguments
Stacktrace:
 [1] error(::String, ::String, ::String) at .\error.jl:42
 [2] sametype_error(::Tuple{HSV{Float64},HSV{Float32}}) at .\promotion.jl:308
 [3] not_sametype(::Tuple{HSV{Float64},HSV{Float32}}, ::Tuple{HSV{Float64},HSV{Float32}}) at .\promotion.jl:302
 [4] promote(::HSV{Float64}, ::HSV{Float32}) at .\promotion.jl:285
 [5] top-level scope at REPL[2]:1
```

A simple (and tactless) solution is adding the rules [here](https://github.com/JuliaGraphics/ColorTypes.jl/blob/7ef745605d6a0897078712a0a7a156c51295b559/src/types.jl#L549-L579).
```julia
    @eval Base.promote_rule(::Type{$Csym{T1}}, ::Type{$Csym{T2}}) where {T1, T2} = $Csym{promote_type(T1, T2)}
    @eval Base.promote_rule(::Type{$acol{T1}}, ::Type{$acol{T2}}) where {T1, T2} = $acol{promote_type(T1, T2)}
    @eval Base.promote_rule(::Type{$cola{T1}}, ::Type{$cola{T2}}) where {T1, T2} = $cola{promote_type(T1, T2)}
```
Of course,
> RGB1 and RGB4 require special handling because of the alphadummy field

However, I don't know what should be returned by `promote_type(HSV, HSV{Float16})`.(`HSV{Float16}`? or `HSV{promote_type(eltype_default(HSV),Float16)}`?)

kimikage:
This fixes #129 and https://github.com/JuliaGraphics/Colors.jl/issues/273.
This PR just focus on fixing broken tests quickly, because I want to release the next version of Colors.jl quickly. So, there is probably a lot of room for improvement.

~This implementation returns `HSV{Float16}` for `promote_type(HSV, HSV{Float16})`. This behavior might be controversial.
This implementation causes `StackOverflow` for unresolved pairs.~
```julia
julia> promote_type(HSV, RGB)
ERROR: StackOverflowError:
```
~I can prevent the `StackOverflow`, but we will end up with an error somewhere else.~
See below.

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -1,12 +1,105 @@
-Base.promote_rule(::Type{T1}, ::Type{T2}) where {T1<:AbstractGray,T2<:AbstractGray} = Gray{promote_type(eltype(T1), eltype(T2))}
-Base.promote_rule(::Type{T1}, ::Type{T2}) where {T1<:AbstractRGB,T2<:AbstractRGB} = RGB{promote_type(eltype(T1), eltype(T2))}
 
-Base.promote_rule(::Type{C3}, ::Type{Cgray}) where {C3<:Color3,Cgray<:AbstractGray} = base_colorant_type(C3){promote_type(eltype(C3), eltype(Cgray))}
-Base.promote_rule(::Type{C3}, ::Type{Cagray}) where {C3<:Color3,Cagray<:AbstractAGray} = alphacolor(base_colorant_type(C3)){promote_type(eltype(C3), eltype(Cagray))}
-Base.promote_rule(::Type{C3}, ::Type{Cgraya}) where {C3<:Color3,Cgraya<:AbstractGrayA} = coloralpha(base_colorant_type(C3)){promote_type(eltype(C3), eltype(Cgraya))}
+function _promote_et(::Type{C1}, ::Type{C2}) where {C1<:Colorant, C2<:Colorant}
+    T1 = isconcretetype(eltype(C1)) ? eltype(C1) : eltypes_supported(C1)
+    T2 = isconcretetype(eltype(C2)) ? eltype(C2) : eltypes_supported(C2)
+    promote_type(T1, T2)
+end
 
-Base.promote_rule(::Type{C3}, ::Type{Cgray}) where {C3<:Transparent3,Cgray<:AbstractGray} = base_colorant_type(C3){promote_type(eltype(C3), eltype(Cgray))}
-Base.promote_rule(::Type{C3}, ::Type{Cgray}) where {C3<:Transparent3,Cgray<:TransparentGray} = base_colorant_type(C3){promote_type(eltype(C3), eltype(Cgray))}
+_with_et(C::UnionAll, et) = isconcretetype(et) ? C{et} : C
+function _transparentcolor_with_et(::Type{C0}, et) where {N, C0<:TransparentColorN{N}}
+    if isconcretetype(et)
+        C0 <: AlphaColor && return AlphaColor{C,et,N} where {C<:Color{et,N-1}}
+        C0 <: ColorAlpha && return ColorAlpha{C,et,N} where {C<:Color{et,N-1}}
+        return TransparentColor{C,et,N} where {C<:Color{et,N-1}}
+    else
+        # Note that the constraint between the element type `T` and the element
+        # type of the base color type `C`.
+        # The element types should be the same, but there is no such a constraint in
+        # the definition of `TransparentColor`/`AlphaColor`/`ColorAlpha`.
+        # The following should be consistent with `base_colorant_type`
+        C0 <: AlphaColor && return AlphaColor{C,T,N} where {T, C<:Color{T,N-1}}
+        C0 <: ColorAlpha && return ColorAlpha{C,T,N} where {T, C<:Color{T,N-1}}
+        return TransparentColor{C,T,N} where {T, C<:Color{T,N-1}}
+    end
+end
+function _transparentcolor_with_et(::Type{C0}, et) where {C0<:TransparentColor}
+    if isconcretetype(et)
+        C0 <: AlphaColor && return AlphaColor{C,et} where {C<:Color{et}}
+        C0 <: ColorAlpha && return ColorAlpha{C,et} where {C<:Color{et}}
+        return TransparentColor{C,et} where {C<:Color{et}}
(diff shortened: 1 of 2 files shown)


timholy:
> I can prevent the StackOverflow, but we will end up with an error somewhere else.

Shouldn't `promote_type(HSV, RGB)` be `Color3`?

timholy:
Actually, what I wrote above is really `typejoin`. See https://github.com/JuliaGraphics/ColorTypes.jl/issues/129#issuecomment-587044704.

kimikage:
I changed the codes, but they are still ugly.:sob:
Now we can get:
```julia
julia> promote_type(HSV, HSV{Float16})
HSV

julia> promote_type(RGB, HSV) === Color3
true
```
I find it difficult to:
- handle the special types `RGB24`, `ARGB32`, `Gray24` and `AGray32`
- handle the `AbstractRGB` and `AbstractGray` (cf. https://github.com/JuliaGraphics/ColorTypes.jl/issues/129#issuecomment-587241714)
- handle the different written forms for transparent color types (e.g. strictness of parameter 
for base color types)
- handle the dispatch order due to `promote_result(T, S, promote_rule(T,S), promote_rule(S,T))`

If you have a good idea, I want to adopt it. If it takes time, I want to focus on fixing the broken promotion rules for concrete types.

src/conversions.jl
```diff
@@ -1,12 +1,88 @@
...
+end
+
+_with_et(C::UnionAll, et) = isconcretetype(et) ? C{et} : C
+function _transparentcolor_with_et(::Type{C0}, et) where {N, C0<:TransparentColorN{N}}
+    if isconcretetype(et)
+        C0 <: AlphaColor && return AlphaColor{C,et,N} where {C<:Color{et,N-1}}
+        C0 <: ColorAlpha && return ColorAlpha{C,et,N} where {C<:Color{et,N-1}}
+        return TransparentColor{C,et,N} where {C<:Color{et,N-1}}
+    else # remove the constraint of eltype `T` of base color type `C`
+        C0 <: AlphaColor && return AlphaColor{C,T,N} where {C<:ColorN{N-1}, T}
```
timholy: Should we use
```julia
AlphaColor{C,T,N} where {T,C<:Color{T,N-1}}
```
Optional; since `C` is unspecified I doubt it matters a lot.

test/conversions.jl
```diff
@@ -114,15 +163,93 @@ end
...
+
+    @test promote_type(HSVA, RGBA) == ColorAlpha{C,T,4} where {T, C<:Color3}
+    @test promote_type(ARGB, AHSV) == AlphaColor{C,T,4} where {T, C<:Color3}
+    @test promote_type(AHSV, RGBA) == Transparent3
+    @test promote_type(ARGB, HSVA) == Transparent3
+
+    @test promote_type(RGB24, HSV) == Color3
+    @test promote_type(RGB24, HSVA{Float16}) == ColorAlpha{C,Float32,4} where C<:Color3{Float32}
+    @test promote_type(ARGB32, HSV) == AlphaColor{C,T,4} where {T, C<:Color3}
+    @test promote_type(ARGB32, HSVA{Float16}) == TransparentColor{C,Float32,4} where C<:Color3{Float32}
```
timholy: Do we need any tests where both are parametric? AFAICT it should be handled fine (these are actually the harder cases), but just wondering. (Optional.)

src/conversions.jl
kimikage: Please excuse my lack of explanation. It looks like a difference of point of view between you and me. (cf. https://github.com/JuliaGraphics/ColorTypes.jl/pull/161#discussion_r379234848)

In fact, the eltype `T` of `TransparentColor` and the eltype of base type `C` are chained (i.e. the same). However, the definitions of `TransparentColor`, `AlphaColor`, and `ColorAlpha` have no such constraints.

This difference is obvious when using "plain" types. However, the extra constraints can be confusing when using the typealiases.
```julia
julia> base_colorant_type(Transparent3) == Transparent3
false
```

We might want to change the type definitions (i.e. introduce the constraints), but I can't imagine the impact.

test/conversions.jl
kimikage: I just forgot because I was so full of non-concrete types.:smile:
I'll add tests.

src/conversions.jl
timholy: Yes, back when ColorTypes was written it wasn't possible to chain them. Now it is, but there's still one thing I want that isn't: chaining the `N` between the Color and the TransparentColor. If it does become possible (https://github.com/JuliaLang/julia/issues/8322), I'd say we should change the type definitions, but I am not sure it's worth it now just to go through it all over again.

kimikage: For consistency with `base_colorant_type`, I added the element type constraint.
https://github.com/JuliaGraphics/ColorTypes.jl/blob/3dfb0045e49d78838d4f530003e43f47484cd630/src/conversions.jl#L15-L22

kimikage:
Note that many promotions with abstract types fail. I will submit another issue report for this. However, for the time being, they may not be fixed.

kimikage:
I am very sorry, but I would like to rethink the promotion rules again.

I would like to implement another rules based on another policy (which is partially similar to [@timholy's proposal](https://github.com/JuliaGraphics/ColorTypes.jl/issues/129#issuecomment-587583341)) and submit a new PR.
And then, let's merge the one we like.

timholy:
No need to apologize; I greatly admire your care and persistence. This kind of systematic reworking takes time but has exceptional long-term benefit to the ecosystem.

kimikage:
I chose the rules implemented in PR #170.

PR #174, opened 2020-03-01
Fix display of strange element types within `colorant_string_with_eltype`

### LINKED ISSUE #151: Problem with `eltype(RGB)` or showing strange "RGB{Any}"

kimikage:
This fixes #151, but does not change `eltype(C)` for the backward compatibility.
This also moves `colorant_string` and `colorant_string_with_eltype` to "show.jl".

### DIFF
--- a/src/show.jl
+++ b/src/show.jl
@@ -6,6 +6,8 @@ show_normed(io::IO, c::Gray24) = print(io, "Gray24(", gray(c), ')')
 show_normed(io::IO, c::RGB24)  = print(io, "RGB24(", red(c), ',', green(c), ',', blue(c), ')')
 show_normed(io::IO, c::ARGB32) = print(io, "ARGB32(", red(c), ',', green(c), ',', blue(c), ',', alpha(c), ')')
 
+# FIXME: handle `Color` and `TransparentColor` correctly
+#       (e.g. `Color{T,4}` such as CMYK is different from `Transparent3`).
 for N = 1:4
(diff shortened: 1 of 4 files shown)


kimikage:
I am willing to release ColorTypes.jl v0.10, but we might have a grace period to the release of ColorTypes.jl v0.10, because of the need to merge some PRs in Colors.jl.

This PR is less important, but may be worth merging as other packages may need to modify their tests.

kimikage:
Thank you for your review.

PR #175, opened 2020-03-01
updates to ones, zeros, oneunit and zero to keep consistency to Base

### LINKED ISSUE #137: support `ones(::Type{<:AbstractGray}, sz)`

johnnychen94:
* make usage like `ones(Gray{N0f8}, 4, 4)` doesn't throw depwarns (closes #137)
* (RFC) make `oneunit(::Gray)` and `zero(::Gray)` valid and keep consistency to Base behavior
* `eltype(zeros(Gray)) == typeof(zero(Gray)) == Gray{N0f8}` (closes https://github.com/JuliaGraphics/Colors.jl/issues/369)

There must be some discussion on these in julialang/julia but I can't find a source...

### DIFF
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -514,7 +514,9 @@ isapprox(x::AbstractGray, y::Number; kwargs...) =
 
 
 zero(::Type{C}) where {C<:Gray} = C(0)
+zero(::C) where {C<:Gray} = C(0)
 oneunit(::Type{C}) where {C<:Gray} = C(1)
+oneunit(::C) where {C<:Gray} = C(1)
 
 function Base.one(::Type{C}) where {C<:Gray}
     Base.depwarn("one($C) will soon switch to returning 1; you might need to switch to `oneunit`", :one)
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -522,6 +522,7 @@ function Base.one(::Type{C}) where {C<:Gray}
     Base.depwarn("one($C) will soon switch to returning 1; you might need to switch to `oneunit`", :one)
     C(1)
 end
+Base.one(::C) where {C<:Gray} = one(C)
 
 Base.broadcastable(x::Colorant) = Ref(x)
 
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -523,5 +523,10 @@ end
 
 Base.broadcastable(x::Colorant) = Ref(x)
 
+# with one(C) deprecated, we need to override the default function call to use oneunit instead of one
+Base.ones(::Type{C}, dims::Tuple{}) where {C<:Gray, N} = fill(oneunit(C))
+Base.ones(::Type{C}, dims::Tuple{Vararg{Integer, N}}) where {C<:Gray, N} = fill(oneunit(C), dims)
+Base.ones(::Type{C}, dims::Union{Integer, AbstractUnitRange}...) where {C<:Gray} = fill(oneunit(C), dims...)
+
 Base.isless(a::AbstractGray, b::AbstractGray) =
     isless(gray(a), gray(b))
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -525,10 +525,19 @@ end
 
 Base.broadcastable(x::Colorant) = Ref(x)
 
-# with one(C) deprecated, we need to override the default function call to use oneunit instead of one
-Base.ones(::Type{C}, dims::Tuple{}) where {C<:Gray, N} = fill(oneunit(C))
-Base.ones(::Type{C}, dims::Tuple{Vararg{Integer, N}}) where {C<:Gray, N} = fill(oneunit(C), dims)
-Base.ones(::Type{C}, dims::Union{Integer, AbstractUnitRange}...) where {C<:Gray} = fill(oneunit(C), dims...)
+# specialize ones and zeros for Gray type:
+#   * Base implementation uses `one` instead of `oneunit`; however, `one(Gray)` will return `1`
+#   * Base implementation use `Array{T}` to initialize the array, however, if we pass
+#     `Gray`, it gives an abstract eltype `Array{Gray}`, which is inconsistent to `zero(Gray)`
+for (fname, felt) in ((:zeros, :zero), (:ones, :oneunit))
+    @eval begin
+        # these dispatches come from Base implementation
+        Base.$fname(::Type{C}, dims::Tuple{}) where {C<:Gray, N} = fill($felt(C))
+        Base.$fname(::Type{C}, dims::NTuple{N, Integer}) where {C<:Gray, N} = fill($felt(C), dims)
+        Base.$fname(::Type{C}, dims::Base.DimOrInd...) where {C<:Gray} = $fname(C, dims)
+        Base.$fname(::Type{C}, dims::NTuple{N, Base.OneTo}) where {C<:Gray, N} = fill($felt(C), map(last, dims))
+    end
+end
 
 Base.isless(a::AbstractGray, b::AbstractGray) =
     isless(gray(a), gray(b))
--- a/test/traits.jl
+++ b/test/traits.jl
@@ -678,13 +678,33 @@ end
     @test oneunit(Gray) === Gray{N0f8}(1)
     @test zero(Gray) === Gray{N0f8}(0)
 
-    for T in (N0f8, Float32)
+    @test ones(Gray) == fill(oneunit(Gray))
+    @test ones(Gray, 4, 4) == fill(oneunit(Gray), 4, 4) == Gray.(ones(N0f8, 4, 4))
+    @test eltype(ones(Gray, 4, 4)) == Gray{N0f8}
+
+    @test zeros(Gray) == fill(zero(Gray))
+    @test zeros(Gray, 4, 4) == fill(zero(Gray), 4, 4) == Gray.(zeros(N0f8, 4, 4))
+    @test eltype(zeros(Gray, 4, 4)) == Gray{N0f8}
+
+    for T in (Bool, N0f8, Float32)
+        for (fname, felt) in ((:zeros, :zero), (:ones, :oneunit))
+            @eval begin
+                @test $fname(Gray{$T}) == fill($felt(Gray{$T}))
+                A = $fname(Gray{$T}, 4, 4)
+                @test A == $fname(Gray{$T}, (4, 4))
+                @test A == $fname(Gray{$T}, Base.OneTo(4), Base.OneTo(4))
+                @test A == $fname(Gray{$T}, (Base.OneTo(4), Base.OneTo(4)))
+                @test eltype(A) == Gray{$T}
+            end
+        end
+
         @test oneunit(Gray{T}) === Gray{T}(1)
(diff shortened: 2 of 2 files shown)


test/traits.jl
```diff
@@ -698,12 +718,4 @@ end
...
     @test_throws MethodError oneunit(RGB{N0f8})
     @test_throws MethodError zero(RGB{N0f8})
-
-    for T in (N0f8, Float32)
-        g = Gray{T}(0.8)
-        @test oneunit(g) == oneunit(T) == Gray(oneunit(T))
-        @test zero(g) == zero(T) == Gray(zero(T))
-    end
-
-    @test_broken one(Gray{Float32}) * g == g * one(Gray{Float32}) == g
```
johnnychen94: ColorVectorSpaces seems already fixes this, so it not a valid test anymore.

test/traits.jl
```diff
@@ -692,9 +699,11 @@ end
...
     @test_throws MethodError zero(RGB{N0f8})
 
-    g = Gray{Float32}(0.8)
-    @test_throws MethodError oneunit(g)
-    @test_throws MethodError zero(g)
+    for T in (N0f8, Float32)
+        g = Gray{T}(0.8)
+        @test oneunit(g) == oneunit(T) == Gray(oneunit(T))
+        @test zero(g) == zero(T) == Gray(zero(T))
+    end
```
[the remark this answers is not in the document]
johnnychen94: I intended to make minimal `git diff`... 😂

johnnychen94:
The latest commit also fixes https://github.com/JuliaGraphics/Colors.jl/issues/369

test/traits.jl
kimikage: The test means `one(Gray)` should return `1`. Do not remove it now.

kimikage:
> not very sure which method would become unnecessary here.

I had assumed that `eltype(ones(Gray)) === Gray`. We will probably need almost all of the added methods. :smile:

test/traits.jl
johnnychen94: Neither `Gray{N0f8}(1) * Gray{N0f8}(0.8)` nor `1 * Gray{N0f8}(0.8` are valid without ColorVectorSpaces...

And `Gray{N0f8} == 1` is true

> The test means one(Gray) should return 1

I'll add one more test for this.

src/traits.jl
```diff
@@ -514,14 +514,31 @@ isapprox(x::AbstractGray, y::Number; kwargs...) =
...
     C(1)
 end
+Base.one(::C) where {C<:Gray} = one(C)
 
 Base.broadcastable(x::Colorant) = Ref(x)
 
+# specialize ones and zeros for Gray type:
+#   * Base implementation uses `one` instead of `oneunit`; however, `one(Gray)` will return `1`
+#   * Base implementation use `Array{T}` to initialize the array, however, if we pass
+#     `Gray`, it gives an abstract eltype `Array{Gray}`, which is inconsistent to `zero(Gray)`
```
kimikage: To be honest, I don't understand what this comment says. I feel the reason is the opposite.
Even though `one(Gray)` should return `1`, not so. Thus, `one` shows the warning. Isn't this PR a workaround to avoid that warning? If `one(Gray)` returns `1`, ones(Gray) returns an array of Gray(1).

Regarding the latter:
```julia
julia> zero(AbstractFloat)
0.0

julia> zeros(AbstractFloat)
0-dimensional Array{AbstractFloat,0}:
0.0
```
`Gray{N0f8}` is just a matter of practicality, i.e. speed or memory efficiency.

johnnychen94: I meant to make `ones(Gray)` still return an array of `Gray(1)` regardless of the fact that `one(Gray) === 1` in the future. Did I misunderstand the original reason for the shift?

I think I need to raise `zeros(AbstractFloat)` issue at julialang/julia.. My idea is to make the array eltype equals to its initializer, which makes more sense to me.

test/traits.jl
kimikage: > Neither Gray{N0f8}(1) * Gray{N0f8}(0.8) nor 1 * Gray{N0f8}(0.8 are valid without ColorVectorSpaces...

You are right. However, I think the broken test makes sense to indicate that "this is broken" and "one is an identity element".

src/traits.jl
kimikage: > I think I need to raise zeros(AbstractFloat) issue at julialang/julia..

Umm... What is the problem?

johnnychen94: That `eltype(ones(AbstractFloat)) === typeof(one(AbstractFloat))` (and for `zeros`)

kimikage: Yes. I understand the claim, but I just want to know the reason.

johnnychen94: Just opened a PR for this https://github.com/JuliaLang/julia/pull/34948

wrt whether `ones(Gray{N0f8}, 4, 4)` should return array of `N0f8` or `Gray{N0f8}`, I'd like to get @timholy involved.

kimikage: Definitely, `ones(Gray{N0f8}, 4, 4)` should not return an array of `N0f8`. Where did the choice come from?

```julia
ones(Gray{N0f8}, 2, 2)
--> Gray{N0f8}[? ?
               ? ?]
--> Gray{N0f8}[one(Gray{N0f8}) one(Gray{N0f8})
               one(Gray{N0f8}) one(Gray{N0f8})]
--> Gray{N0f8}[1 1
               1 1]
--> Gray{N0f8}[Gray{N0f8}(1) Gray{N0f8}(1) 
               Gray{N0f8}(1) Gray{N0f8}(1)]
```

timholy:
`zeros` should be fine but that already seems to work properly.

`ones` is more problematic in that I don't think `ones` should return anything different from an array of `one(T)` but the distinction between `one` and `oneunit` causes trouble. xref https://github.com/JuliaGraphics/Colors.jl/issues/369#issuecomment-593121207 and https://github.com/JuliaLang/julia/pull/34948#issuecomment-593120343.

johnnychen94:
Since we have different rules for `Gray` and `Number`, wouldn't that be more surprising when `ones` returns an array of `Number` while its siblings `zeros` returns an array of `Gray`?

Also, what benefits would `one(Gray)` be by saying it returns a multiplicative identity for `Gray` when we disallow the multiplication...except that users don't get MethodError when they call `ones`?

I can understand the general rules here but it looks like I'm still keeping confused when using these in the wild; it's very hard to not thinking `zeros`/`ones` as the counterpart of that in Matlab and NumPy worlds.

Said this, I would feel a lot better if `one` returns the same type as `zero`, whatever it might be. (For Base consistency or for general image processing convenience)

johnnychen94:
Let's not include this PR in v0.10 and leave more time to think about it.

timholy:
The point is that if `one` is supposed to be the multiplicative identity, it needs to at least be that. So we have to decide whether `Gray(1)*Gray(1) == Gray(1)`. Currently it's not unless you load ColorVectorSpace. And for RGB the situation is even more confusing.

Really, `ones` should just die and be replaced by `fill`.

johnnychen94:
I believe it's harmless to support `0.8 * Gray(1.)` and `Gray(1.) * 0.8` here, which can be read as "multiply the grayscale by 0.8"? (but not `Gray(1) * Gray(1)`). This makes `1` a valid multiplicative identity.

> Really, ones should just die and be replaced by fill.

This becomes the most unambiguous solution to me.

timholy:
> This makes 1 a valid multiplicative identity for Gray

If we support multiplication here at all. From a standpoint of pure colorimetry I'm not certain that multiplying RGB by anything other than 1 makes sense. (Of course we have to support it in image processing...)

kimikage:
> I don't think ones should return anything different from an array of `one(T)`

_Originally posted by @timholy in https://github.com/JuliaGraphics/ColorTypes.jl/pull/175#issuecomment-593123622_

I think `ones(T, ...)` should return an array of the first argument `T`, because `ones` (and `zeros`) should be for the "array constructor".
If `one(T)` should not be converted to `T` (e.g. `Dates.Day`), `ones` should throw an error. However, 
both a  `Gray` value and `1` can be converted to `Gray`. 
There is disagreement about whether that is appropriate, but at least, the current implementation agrees with that (cf. https://github.com/JuliaGraphics/ColorTypes.jl/pull/175#discussion_r386110428).

Whether `ones(Gray, ...)` should return an array of `Gray{N0f8}` is another matter. I don't care about it.

timholy:
I'm converging on the following strategy:
1. delete `one(Gray{T})` from this package altogether (`one(Gray{T})` will throw an error if you've only loaded ColorTypes)
2. In ColorVectorSpace, define `one(Gray{T})` as `Gray{T}(1)`. That's the opposite of what our depwarn here indicates, but we'd be using the longstanding existence of the depwarn as an excuse for step 1 above.
3. Move this PR to ColorVectorSpace

kimikage:
It looks reasonable to me. I'm sure some people will wonder why ColorTypes doesn't define `one`, though. :smile:

timholy:
I am beginning to think we need to extend Julia's printing of `MethodError`s to allow packages to push custom hints. Similar to https://github.com/JuliaLang/julia/pull/34642 and many similar changes, but package-extensible.

kimikage:
The issues related to this PR have been settled by PR #243.

Now, `ones(Gray, ...)`, `zeros(RGB, ...)` etc. are not recommended for the performance reason.

```julia
julia> ones(Gray{N0f8}, 4, 4)
4×4 Matrix{Gray{N0f8}}:
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0
 1.0  1.0  1.0  1.0

julia> ones(Gray, 4, 4)
4×4 Matrix{Gray}:
 Gray{N0f8}(1.0)  Gray{N0f8}(1.0)  Gray{N0f8}(1.0)  Gray{N0f8}(1.0)
 Gray{N0f8}(1.0)  Gray{N0f8}(1.0)  Gray{N0f8}(1.0)  Gray{N0f8}(1.0)
 Gray{N0f8}(1.0)  Gray{N0f8}(1.0)  Gray{N0f8}(1.0)  Gray{N0f8}(1.0)
 Gray{N0f8}(1.0)  Gray{N0f8}(1.0)  Gray{N0f8}(1.0)  Gray{N0f8}(1.0)
```

PR #177, opened 2020-03-18
Support alpha(::Number) and ensure alpha/gray return compatible types

### LINKED ISSUE #162: Is it necessary to add alpha(x::Number)?
In some cases, ColorTypes treats scalar numbers like grays.
```julia
julia> Gray(0.5) == 0.5
true
```
The existence of `gray(x::Number)` is another example of that.
https://github.com/JuliaGraphics/ColorTypes.jl/blob/5b2e50d62b8fd47a860c2e0543446144d8627c87/src/traits.jl#L36

According to that idea, perhaps we need `alpha(x::Number)`:
```julia
alpha(x::Number) = alpha(Gray(x))
```
or
```julia
alpha(x::Number) = oneunit(x)
```

timholy:
Closes #162. I decided to continue to support any `::Number` as long as it can be `convert`ed to a supported type. Consequently
```julia
julia> using FixedPointNumbers, ColorTypes

julia> FixedPointNumbers.floattype(::Type{Complex{T}}) where T = Complex{floattype(T)}

julia> gray(1+0im)
1.0N0f8

julia> gray(1+1im)
ERROR: InexactError: Int64(1 + 1im)
Stacktrace:
 [1] Real at ./complex.jl:37 [inlined]
 [2] convert at ./number.jl:7 [inlined]
 [3] FixedPoint at /home/tim/.julia/packages/FixedPointNumbers/w2pxG/src/FixedPointNumbers.jl:56 [inlined]
 [4] convert at ./number.jl:7 [inlined]
 [5] gray(::Complex{Int64}) at /home/tim/.julia/dev/ColorTypes/src/traits.jl:37
 [6] top-level scope at REPL[5]:1
```
(Not sure if that is worth adding to the tests.) This is consistent with the general philosophy that types are about dispatch and algorithmic efficiency, but that functionality should depend primarily on value.

Breaking change: yes (conversions that can't be performed will now error, and some return values will be of different types).

### DIFF
--- a/test/types.jl
+++ b/test/types.jl
@@ -266,14 +266,8 @@ end
         @test GrayA{N0f16}(val) === GrayA{N0f16}(0.2, 1)
         @test AGray32(val) === AGray32(0.2, 1)
         @test AGray32(val, 0.8) === AGray32(0.2, 0.8)
-        if val isa FixedPoint
-            @test AGray(val, 1) === AGray{Float32}(0.2, 1) # inconsistent eltype
-            @test_broken @inferred(AGray(val, 1)) === AGray{T}(0.2, 1)
-            @test_broken @inferred(GrayA(val, 0)) === GrayA{T}(0.2, 0)
-        else
-            @test @inferred(AGray(val, 1)) === AGray{T}(0.2, 1)
-            @test @inferred(GrayA(val, 0)) === GrayA{T}(0.2, 0)
-        end
+        @test @inferred(AGray(val, 1)) === AGray{T}(0.2, 1)
+        @test @inferred(GrayA(val, 0)) === GrayA{T}(0.2, 0)
         Ta = val isa AbstractGray ? T : Float64
         if val isa Gray24
             @test_broken @inferred(AGray(val, 0.8)) === AGray{Ta}(val, 0.8)
--- a/test/types.jl
+++ b/test/types.jl
@@ -145,8 +145,8 @@ end
                 @test C{N0f16}(val1,val2,val1,0.8) === C(0.2N0f16,0.6N0f16,0.2N0f16,0.8N0f16)
             end
             # 1-arg constructor
-            @test_broken C(val1) === C{typeof(val1)}(0.2,0.2,0.2,1)
-            @test_broken C{N0f8}(val1) === C{N0f8}(0.2,0.2,0.2,1)
+            @test C(val1) === C{typeof(val1)}(0.2,0.2,0.2,1)
+            @test C{N0f8}(val1) === C{N0f8}(0.2,0.2,0.2,1)
         end
 
         @test_throws ArgumentError C(2,1,0) # integers
@@ -161,6 +161,7 @@ end
             @test isa(C(val,val,val), C)
             @test isa(C(val,val,val,0.8), C)
             @test isa(C(val,val,val,val), C) # no exception thrown
+            @test alpha(C(val)) === oneunit(val)
         end
     end
 end
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -7,33 +7,33 @@ without an alpha channel, it will always return 1.
 alpha(c::TransparentColor) = c.alpha
 alpha(c::Color)   = oneunit(eltype(c))
 alpha(c::RGB24)   = N0f8(1)
-alpha(c::ARGB32)  = N0f8((c.color & 0xff000000)>>24, 0)
-alpha(c::AGray32) = N0f8((c.color & 0xff000000)>>24, 0)
+alpha(c::ARGB32)  = reinterpret(N0f8, (c.color >> 0x18) % UInt8)
+alpha(c::AGray32) = reinterpret(N0f8, (c.color >> 0x18) % UInt8)
 alpha(x::Number)  = convert(eltype(ccolor(Gray, typeof(x))), oneunit(x))  # ensures it's a type supported by Gray (which has widest eltype support)
 
 "`red(c)` returns the red component of an `AbstractRGB` opaque or transparent color."
 red(c::AbstractRGB   ) = c.r
 red(c::TransparentRGB) = c.r
-red(c::RGB24)  = N0f8((c.color & 0x00ff0000)>>16, 0)
-red(c::ARGB32) = N0f8((c.color & 0x00ff0000)>>16, 0)
+red(c::RGB24)  = reinterpret(N0f8, (c.color >> 0x10) % UInt8)
+red(c::ARGB32) = reinterpret(N0f8, (c.color >> 0x10) % UInt8)
 
 "`green(c)` returns the green component of an `AbstractRGB` opaque or transparent color."
 green(c::AbstractRGB   ) = c.g
 green(c::TransparentRGB) = c.g
-green(c::RGB24)  = N0f8((c.color & 0x0000ff00)>>8, 0)
-green(c::ARGB32) = N0f8((c.color & 0x0000ff00)>>8, 0)
+green(c::RGB24)  = reinterpret(N0f8, (c.color >> 0x08) % UInt8)
+green(c::ARGB32) = reinterpret(N0f8, (c.color >> 0x08) % UInt8)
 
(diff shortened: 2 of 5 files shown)


kimikage:
This is off-topic and optional, but see also https://github.com/JuliaGraphics/ColorTypes.jl/issues/162#issuecomment-586131860

timholy:
I added that. There's no performance improvement with these new implementations, but I agree it's a better approach.

I also noticed an issue this caused with respect to alpha-constructors, but which was not tested in this package. I caught it by working on ColorVectorSpace; it makes sense to leave this open until that work comes to a conclusion.

kimikage:
If we fix #156 in the near future, I think there are roughly two ways.
1. specializing the constructor with `Normed` argument
2. promoting the eltype to `Float`

Although I think the former is simpler, we may change the alpha-constructors again.

kimikage:
Oops. This seems to change the behavior of constructor with `GrayLike` arguments in "rare" cases.
https://github.com/JuliaGraphics/ColorTypes.jl/blob/4bb8aa9c918c376a14c2717c4e9edef67224f94c/src/types.jl#L600-L607
I think `real()` is more appropriate there than `gray()` semantically.

Changing the `ARGB32` constructor (cf. https://github.com/JuliaGraphics/ColorTypes.jl/pull/183#issuecomment-616958191) needs to add the `ARGB32` constructor with `GrayLike` arguments for the backward compatibility (see below).

Therefore, we need to plan on which PR to make the changes to `real()`, and the merging order.

WRT `GrayLike` arguments, only opaque RGB constructors support them.
```julia
julia> ARGB32(1, 1, 1, Gray(0.5)) # a special case due to the side-effect of inconsistent implementation
ARGB32(1.0N0f8,1.0N0f8,1.0N0f8,0.502N0f8)

julia> ARGB(1, 1, 1, Gray(0.5))
ERROR: MethodError: no method matching ARGB(::Int64, ::Int64, ::Int64, ::Gray{Float64})

julia> AGray{Float32}(1, Gray(0.5))
ERROR: MethodError: no method matching AGray{Float32}(::Int64, ::Gray{Float64})
```
I don't like to confuse `Gray` with `Real`, so I'm not going to enhance the `GrayLike` support. However, it might be good to discuss that in a separate issue report.

test/types.jl
```diff
@@ -236,6 +237,18 @@ end
...
+    for x in (1, N0f8(0.5), 0.5f0, 0.5)
+        T = isa(x, Int) ? N0f8 : typeof(x)
+        @test @inferred(AGray(x)) === @inferred(AGray(x, 1)) === AGray{T}(x, 1)
+        @test @inferred(GrayA(x)) === @inferred(GrayA(x, 1)) === GrayA{T}(x, 1)
+        g = Gray(x)
+        @test @inferred(AGray(g)) === @inferred(AGray(g, 1)) === AGray{T}(x, 1)
+        @test @inferred(GrayA(g)) === @inferred(GrayA(g, 1)) === GrayA{T}(x, 1)
+    end
+    @test AGray32(Gray(0.5)) === AGray32(Gray(0.5), 1) === AGray32(N0f8(0.5)) === AGray32(N0f8(0.5), 1) === reinterpret(AGray32, 0xff808080)
+end
```
kimikage, line 250: What is the purpose of adding these?

timholy: Without the change to `src/types.jl` above, `x = N0f8(0.5); AGray(x, 1)` returns an `AGray{Float32}` because `promote(x, 1)` chooses `Float32`. But with an integer argument we know the only valid settings of `alpha` are 0 and 1, so a valid implementation would be to convert it a `Bool` first (drawing the error if that fails) and then call `promote`. But note that `promote(x, true)` chooses `N0f8`, so the final type produced is strongly dependent on internal details of the implementation of the constructor.

To me it seems that a more agnostic approach is to first convert `alpha` to a supported eltype and then do the promotion. Note that changing the signature of those same to methods to `alpha::GrayLike` would circumvent the error you noted in https://github.com/JuliaGraphics/ColorTypes.jl/pull/177#issuecomment-617597990. Though I don't feel strongly that we need to support `Gray` values for the alpha channel, making it consistently throw an error would also be OK.

kimikage: I finally got the purpose. :+1:  (For such a purpose, I prefer to write the types extensionally, though.)

> But with an integer argument we know the only valid settings of `alpha` are 0 and 1

I agree for the most part. However, so far, we have not checked the range of valid values ​​for each "color" component. Actually , we only check the valid range of `Normed` "type", so `AGray{N1f7}(1, 2)` does not throw errors.
Therefore, I prefer the more agnostic approach.

kimikage: > I don't feel strongly that we need to support `Gray` values for the alpha channel,

I agree. However, it is not only problem with the alpha channel.
```julia
julia> RGB{N0f8}(Gray(1), 1, 1)
RGB{N0f8}(1.0,1.0,1.0)

julia> ARGB{N0f8}(Gray(1), 1, 1, 0.5)
ERROR: MethodError: no method matching ARGB{Normed{UInt8,8}}(::Gray{Normed{UInt8,8}}, ::Int64, ::Int64, ::Float64)
```
"Unfortunately", the possible way to make them consistent may be to enhance the support for `GrayLike`, and I support the status quo. :sweat_smile:

timholy: I do think it's reasonable to construct an RGB from gray channels, as it is the inverse operation of color separation.

timholy:
This is now being proposed for merging.

kimikage:
I don't think there is any problem with this PR, but I am thinking of a radical overhaul of the constructors (cf. #197).

I would like to merge PRs with less impacts first.

timholy:
> I would like to merge PRs with less impacts first.

Can you clarify your intent? (What are "less impacts"?) Meaning, should this PR wait on #197 or other open PRs or are you saying this should merge before #197?

kimikage:
I meant #183, #187 and (something like) #189. The PR #197 is speculative.

timholy:
Sorry I didn't merge #183. Done now.

kimikage:
WRT the problem with the eltype promotion (https://github.com/JuliaGraphics/ColorTypes.jl/pull/177#discussion_r412810164), I think the following rule (which does not explicitly depend on `N0f8`) is better.
```julia
# the modified version of `promote_eltype()`
@inline function promote_args_type(::Type{C}, alpha::Ta, vals...) where {Ta, C <: TransparentColor}
    T = promote_type(map(typeof, vals)...)
    if T <: Union{Integer, FixedPoint} && Ta <: Integer
        _promote_args_type(eltype_ub(C), eltype_default(C), T)
    else
        _promote_args_type(eltype_ub(C), eltype_default(C), promote_type(T, Ta))
    end
end
```

However, it is difficult to apply the rule consistently in the current constructor methods.Therefore, I think `gray(alpha)` is a reasonable solution in this PR.

kimikage:
This is my oversight, but some tests should be added into "conversions from/to real numbers", "transparent rgb constructors" and so on, as mentioned in https://github.com/JuliaGraphics/ColorTypes.jl/issues/162#issuecomment-649298731, 

Although the problem is rooted in the lack of consistency in the constructor design, but we need some regression tests to improve the constructors. Of course, we can also add the tests in a separate PR.

timholy:
Sure, wherever you want to add the tests is fine with me. Since this PR is late in the scheduled-merge queue, don't worry about conflicts, we can rebase when we get back to this.

kimikage:
FixedPointNumbers v0.8.2, which includes https://github.com/JuliaMath/FixedPointNumbers.jl/pull/177, was registered.

kimikage:
FYI, I rewrote the tests for this in #240, so there is a conflict. I think it would be easier to just cherry-pick the changes in src and rewrite the tests.

kimikage:
I ran `rebase` and updated the tests.

kimikage:
Although the relaxation to `Gray{T<:Real}` requires reconsideration of the behavior of `gray`, I think this behavior is fine for v0.12.

kimikage:
Note that the behavior of the constructors will be changed again in PR #197. It is not recommended to merge the fixes for this PR's changes into the main branches of the downstream packages.

PR #178, opened 2020-04-06
add custom error hints for zero and one to ColorVectorSpace

johnnychen94:
This gives a user-friendly error message when ColorVectorSpace isn't loaded.

```julia
julia> zero(RGB)
ERROR: MethodError: no method matching zero(::Type{RGB})
did you forget to `using ColorVectorSpace`?
Closest candidates are:
...

julia> zero(RGB(1, 1, 1))
ERROR: MethodError: no method matching zero(::RGB{FixedPointNumbers.Normed{UInt8,8}})
did you forget to `using ColorVectorSpace`?
Closest candidates are:
...
```

### DIFF
--- a/src/ColorTypes.jl
+++ b/src/ColorTypes.jl
@@ -85,4 +85,8 @@ if VERSION >= v"1.1" # work around https://github.com/JuliaLang/julia/issues/341
     _precompile_()
 end
 
+function __init__()
+    include(joinpath(@__DIR__, "error_hints.jl"))
+end
+
 end # module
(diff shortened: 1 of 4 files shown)


timholy:
Nice! Instead of "Did you forget..." how about "You may need to..."?

johnnychen94:
Didn't bump a patch version for this since it's a cutting-edge new feature.

johnnychen94:
Looks like this PR breaks the downstream modules during precompilation:

https://github.com/JuliaImages/Images.jl/runs/593248480

timholy:
Rats. Probably best is to emit the registrations directly from the `__init__`. E.g.,

src/ColorTypes.jl
```julia
function register_hints()
    if VERSION >= ...
    end
end

__init__() = register_hints()
```
Can I leave this up to you? I have a pressing deadline.

johnnychen94:
Not sure if I understand you correctly, but I can try it later today.

PR #181, opened 2020-04-15
Add setup block to doctests

timholy:
This makes them standalone, which is necessary for juliaimages docs. Alternatively we could use the same `setdocmeta!` statement there, but this seems to have lower overhead for any other package that might want to incorporate these docs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ os:
   - osx
 julia:
   - 1.0
-  - 1.3
+  - 1
   - nightly
 notifications:
   email: false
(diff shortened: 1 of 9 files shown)


kimikage:
> this seems to have lower overhead

I'm just curious whether the overhead is fatal. 
`doctest(source, modules; kwargs...)` might be helpful but I have never used it.
https://juliadocs.github.io/Documenter.jl/stable/lib/public/#Documenter.doctest

timholy:
To clarify, I meant "maintenance overhead" (something you have to add to fix an error) rather than "performance overhead." Meaning that if we put the statements in `setup` inline statements, then any package that wants to reexport their docstrings doesn't have to worry about setup.

PR #182, opened 2020-04-16
fixes a precompilation error when registering error hints

johnnychen94:
#178 is buggy (see https://github.com/JuliaGraphics/ColorTypes.jl/pull/178#issuecomment-614902798), this PR fixes that

Bump a version because it already breaks CI for downstream packages, e.g., https://github.com/JuliaGraphics/ColorSchemes.jl/pull/33, https://github.com/JuliaImages/Images.jl/runs/593248480

~To make sure this kind of issue doesn't happen again, I also added Colors and ImageCore tests in CI.~

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ColorTypes"
 uuid = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
-version = "0.10.1"
+version = "0.10.2"
 
 [deps]
 FixedPointNumbers = "53c48c17-4a7d-5ca2-90c5-79b7896eea93"
--- a/src/ColorTypes.jl
+++ b/src/ColorTypes.jl
@@ -45,6 +45,7 @@ include("traits.jl")
 include("conversions.jl")
 include("show.jl")
 include("operations.jl")
+include("error_hints.jl")
 
 Base.@deprecate_binding RGB1 XRGB
 Base.@deprecate_binding RGB4 RGBX
@@ -85,8 +86,6 @@ if VERSION >= v"1.1" # work around https://github.com/JuliaLang/julia/issues/341
     _precompile_()
 end
 
-function __init__()
-    include(joinpath(@__DIR__, "error_hints.jl"))
-end
+__init__() = register_hints()
 
 end # module
--- a/src/error_hints.jl
+++ b/src/error_hints.jl
@@ -1,11 +1,13 @@
 # provided by https://github.com/JuliaLang/julia/pull/35094
-if VERSION >= v"1.5.0-DEV.491"
-    register_error_hint(MethodError) do io, exc, argtypes, kwargs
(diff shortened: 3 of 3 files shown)


.travis.yml
```diff
@@ -1,12 +1,18 @@
...
-  - 1
   - nightly
 notifications:
   email: false
+
+script:
+  - julia --project --color=yes -e 'using Pkg;
+                                    Pkg.instantiate();
+                                    Pkg.test(coverage=true);
+                                    Pkg.add(["Colors", "ImageCore"]);
```
johnnychen94, line 13: 😕 I'm really confused about why it errors in CI but not locally.

https://travis-ci.org/github/JuliaGraphics/ColorTypes.jl/builds/676029217#L269

kimikage:
> To make sure this kind of issue doesn't happen again, I also added Colors and ImageCore tests in CI.

Is this a permanent measure? I don't really agree to make bidirectional edges of dependency (responsibility). :confused:
These packages are maintained by mainly the same members, though.:sweat_smile:

.travis.yml
kimikage: Try deleting Manifest.toml before running the test locally.
The CI runs in the environment of "ColorTypes" Project.toml. Since the environment is just for ColorTypes itself, Manifest.toml does not have the "ColorTypes" entry.

johnnychen94: This is how I did locally, which I believe is a clean build.

```bash
git clone https://github.com/JuliaGraphics/ColorTypes.jl.git
cd ColorTypes.jl
JULIA_DEPOT_PATH=~/julia_temp julia --project=. --color=yes -e 'using Pkg; \
                                    Pkg.instantiate(); \
                                    Pkg.test(coverage=true); \
                                    Pkg.add(["Colors", "ImageCore"]);
```

johnnychen94:
Sigh... I failed to set up the CI environment so I reverted those commits.

I've tested it locally that it fixed the issue in #178

.travis.yml
kimikage: What version of Julia are you using? Pkg.jl has changed at least in the past month.

cormullion:
Is [this Travis failure](https://github.com/JuliaGraphics/ColorSchemes.jl/pull/33) the same as this?

johnnychen94:
Really sorry for the trouble... I didn't know this `include` is an issue.

timholy:
No worries, I didn't either or I would have caught this when I reviewed it.

.travis.yml
kimikage: cf https://github.com/JuliaLang/Pkg.jl/pull/1696

Adding Colors to the ColorTypes project causes a cyclic dependency, although it is only a test dependency.

johnnychen94: Correct, I can now reproduce it with Julia 1.5-DEV but not with Julia 1.4. Thanks for the link.

PR #171, opened 2020-02-23
Fix RGB24/ARGB32 conversions (Fixes #163)

kimikage:
This fixes #163.

This PR also moves `real` and `reinterpret` to "conversions.jl", and removes the methods which were once needed due to ambiguity.

PR #183, opened 2020-04-19
Small improvements to docstrings and error messages

timholy:


src/traits.jl
```diff
@@ -303,33 +303,34 @@ parametric_colorant(::Type{ARGB32}) = ARGB{N0f8}
...
-    floattype(::Type{T}) where T<:Colorant
+    CF = floattype(C::Type)
 
-Promote storage data type of colorant `T` to `AbstractFloat` while keep the
+Promote storage data type of colorant type `C` to `AbstractFloat` while keeping the
 `base_colorant_type` the same.
 
 !!! info
 
     Non-parametric colorants will be promote to corresponding parametric
```
johnnychen94, line 313: My legacy mistake 😂 
```suggestion
    Non-parametric colorants will be promoted to corresponding parametric
```

kimikage:
I wonder why I didn't notice that in https://github.com/JuliaGraphics/ColorTypes.jl/pull/171, but why does not `ARGB32()` use `checkval` and `throw_colorerror`?

https://github.com/JuliaGraphics/ColorTypes.jl/blob/4bb8aa9c918c376a14c2717c4e9edef67224f94c/src/types.jl#L347

`RGB24()` uses `checkval` and `_rem`.
https://github.com/JuliaGraphics/ColorTypes.jl/blob/4bb8aa9c918c376a14c2717c4e9edef67224f94c/src/types.jl#L323-L326

FYI, I made `N0f8(number)` and `number % N0f8` slightly different.(cf. https://github.com/JuliaMath/FixedPointNumbers.jl/pull/131)
So,
```julia
julia> RGB24(-0.0019, 0, 0)
RGB24(0.0N0f8,0.0N0f8,0.0N0f8)

julia> ARGB32(-0.0019, 0, 0)
ERROR: ArgumentError: Normed{UInt8,8} is an 8-bit type representing 256 values from 0.0 to 1.0; cannot represent -0.0019
```
:confused:

This issue hasn't become apparent in the last six months. (Strictly speaking, I should have seen this error message when I submitted https://github.com/JuliaGraphics/Colors.jl/pull/408, but I didn't notice the difference.:sweat:)
So I think the impact of the change is small but will be breaking change.

timholy:
> why does not ARGB32() use checkval and throw_colorerror

Probably just an omission. I'd certainly favor consistency.

kimikage:
`checkval` and `throw_colorerror` are going to be changed in PR #187. So, are the following steps OK?
1. Merge this PR (#183)
2. Add the change of `ARGB32` constructor to PR #187
3. Merge PR #187

Or, do we need cherry picking between these PRs?

In addition, by changing the first argument of `throw_colorerror`, we can read the user's mind a little more.:crystal_ball:
```julia
julia> RGB24(0x112233)
ERROR: ArgumentError: element type Normed{UInt8,8} is an 8-bit type representing 256 values from 0.0 to 1.0,
  but the values (0x00112233, 0x00112233, 0x00112233) do not lie within this range.
```

kimikage:
I want to merge the minor fixes before the fixes related to the invalidation. (Perhaps some of the fixes of invalidations will not involve public API changes, but I think it is better to gather them into a new minor version.)

@timholy, is this ready to be merged?

Issue #167, opened 2020-02-21
Distinguish meaning of fields, e.g. straight alpha and pre-multiplied alpha


goretkin:
The example in https://github.com/JuliaGraphics/ColorTypes.jl/issues/166 made me wonder whether `ColorTypes.jl` should try to encode the difference between e.g. pre-multiplied alpha and straight alpha. This kind of semantic information would be necessary to prevent the sort of Hungarian method notation below when trying to implement an alpha compositing operation.

 (pma and sa for premultiplied alpha and straight alpha):

```julia
function over_pma(top::TransparentColor, bot::TransparentColor)
  out_T = promote_type(typeof(bot), typeof(top))
  @show out_T
  top_α = alpha(top)
  bot_α = alpha(bot) * (1 - top_α)
  # assume RGB is premultiplied alpha
  out_α = top_α + bot_α
  out_c = RGB(top) + RGB(bot) * (1 - top_α)
  out_T(out_c, out_α)
end

over_sa(top::TransparentColor, bot::TransparentColor) = over_pma(sa_to_pma(top), sa_to_pma(bot))
```

This could be taken to an extreme, where there is annotation to know whether a color represents, I don't know, transmissivity or emission. I think the two types of interpretation of color + alpha is a more practical matter, as different APIs use different formats: https://microsoft.github.io/Win2D/html/PremultipliedAlpha.htm .

kimikage:
Whether the difference should be handled in the "type" or in the "function" (method) is controversial.
IMO, "ideally", we should create new abstract types like `PMAColor` and `PMColorA` under `TransparentColor`.
However, I doubt that "concrete" types need to be defined in ColorTypes.jl. In particular, defining the constructor is an issue.

The other thing we have to consider is whether alpha compositing is the responsibility of Colors.jl.

Also, be careful what `colorant"transparent"` should return. In [CSS Color Level 3](https://www.w3.org/TR/css-color-3/#transparent), it means a transparent "black" (i.e. `colorant"rgba(0,0,0,0)"`). This can be a problem with gradations, and in fact, different web browsers behave differently. PMA can help resolve the gradation problem, but it should affect many packages.

kimikage:
In addition, the PMA is not fully compatible with cylindrical spaces, i.e. color types with hue (e.g. HSV, LCHab). Although there is no need to prohibit users from using such cylindrical PMA colors, we need to consider what ColorTypes.jl should provide.

[...]

Issue #115, opened 2018-12-15
Implement RGB8 / RGB332


scls19fr:
Similar to #114 and following https://github.com/JuliaGraphics/Colors.jl/issues/332#issuecomment-447551891

Implementing RGB8 [8-bit colors](https://en.wikipedia.org/wiki/8-bit_color) which are generally represented as

```
7 6 5|4 3 2|1 0
-----|-----|---
R R R|G G G|B B
```

* 3 bits for **red** channel (8 values)
* 3 bits for **green** channel (8 values)
* 2 bits for **blue** channel (4 values)

might be considered

scls19fr:
WIP:

```julia
import ColorTypes: blue, green, red
import Base: UInt8

"""
`RGB8` is Red-Green-Blue colorspace represented by one byte (8 bits).

7  6  5  4  3  2  1  0
R  R  R  G  G  G  B  B
"""
struct RGB8
    value::UInt8
end

function RGB8(r, g, b)
    RGB8(r << 5 + g << 2 + b)
end

UInt8(c::RGB8) = c.value

blue(c::RGB8) = c.value & 0x03
green(c::RGB8) = (c.value & 0x1c) >> 2
red(c::RGB8) = (c.value & 0xe0) >> 5

using Test
@testset "RGB8" begin
    c = RGB8(7, 7, 3)  # white
    @test red(c) == 7
    @test green(c) == 7
    @test blue(c) == 3
    @test UInt8(c) == 255

    c = RGB8(0, 0, 0)  # black
    @test UInt8(c) == 0

    c = RGB8(7, 0, 0)  # red
    @test UInt8(c) == 224

    c = RGB8(0, 7, 0)  # green
    @test UInt8(c) == 28

    c = RGB8(0, 0, 3)  # blue
    @test UInt8(c) == 3

end
```

I will submit a PR for `RGB8` after #116 being reviewed and fixed

kimikage:
I think the name of `RGB8` is ambiguous.  (cf. https://github.com/JuliaGraphics/ColorTypes.jl/issues/114#issuecomment-465434197)

Sometimes the RGB8 means the RGB888, i.e. `RGB24`.   
The thing that is more worrisome to me is the `RGB4` historically represents a 32-bit color.

scls19fr:
Maybe we should consider naming it `RGB332`

timholy:
[...]

PR #184, opened 2020-04-19
RFC: `eltype`->`eltypec`, `length`->`lengthc`

timholy:
Long ago this package decided to treat colorants as "scalars" and *not* support Julia's standard iteration interface over color channels. We implemented `mapc`, `reducec`, and `mapreducec` to fill the need to sometimes write code that iterates over channels. By and large, Colorants have mostly been thought of similar to how one would typically work with numbers, albeit with some caution to address subtleties of color math (aka, ColorVectorSpace).

However, there are violations of the standard number iterface: for numbers, typically we expect `eltype(T) == T` and `length(x) == 1`, neither of which were true for some Colorants.  Likewise, the core iterator trait `Base.IteratorSize` returns `HasLength()` for `Colorant`s but `HasShape{0}()` for numbers. Finally, `iterate(c::Colorant)` currently throws an error, but `iterate(2)` does not. (Numbers are iterable in Julia to support generic code for which you can pass either `dims = (2,3,5)`, `dims = [2,3,5]`, or `dims = 2`.)

This testing-the-waters PR takes a disruptive step towards treating Colorants just like numbers with respect to the entire iterator interface. It deprecates `eltype(C)` in favor of `eltypec(C)`, and `length(c)` in favor of `lengthc(c)`. To complete the transition, after a suitable period we might consider returning `eltype(C) === C`, `length(c) == 1`, and support iteration of the same style as for Numbers. (Alternatively, `length` and `iterate` could both throw a `MethodError`.)

Needless to say, this has the potential for a lot of downstream code churn. An alternative is to go completely the other way and treat colorants as containers for their color channels. That too would be quite disruptive, perhaps more disruptive than this PR. The third option, of course, is the status quo, which works surprisingly well but still results in occasional "huh??" outcomes.

This package is used by 127 registered packages but only a few devs watch it, so after giving watchers first crack at this PR we should probably CC some additional folks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,9 +368,9 @@ must be identical.  The `AA` is ignored, but in the corresponding
 One of the nicest things about this package is that it provides a rich
 set of trait-functions for working with color types:
 
-- `eltype(c)` extracts the underlying element type, e.g., `Float32`
+- `eltypec(c)` extracts the underlying element type, e.g., `Float32`
 
-- `length(c)` extracts the number of components (including `alpha`, if present)
+- `lengthc(c)` extracts the number of components (including `alpha`, if present)
 
 - `alphacolor(c)` and `coloralpha(c)` convert a `Color` to an object
   with transparency (either `ARGB` or `RGBA`, respectively).
--- a/src/ColorTypes.jl
+++ b/src/ColorTypes.jl
@@ -6,7 +6,7 @@ using Base: @pure
 
 const Fractional = Union{AbstractFloat, FixedPoint}
 
-import Base: ==, isapprox, hash, convert, eltype, length, show, oneunit, zero, reinterpret, getindex
+import Base: ==, isapprox, hash, convert, show, oneunit, zero, reinterpret, getindex
 using Random
 import Random: rand
 
@@ -33,6 +33,7 @@ export RGB24, ARGB32, Gray24, AGray32
 
 
 ## Functions
+export eltypec, lengthc
 export base_color_type, base_colorant_type, ccolor, color, color_type, parametric_colorant
 export alphacolor, coloralpha
 export alpha, red, green, blue, gray   # accessor functions that generalize to RGB24, etc.
@@ -49,6 +50,12 @@ include("error_hints.jl")
 
 Base.@deprecate_binding RGB1 XRGB
 Base.@deprecate_binding RGB4 RGBX
+
+import Base: eltype, length
+@deprecate eltype(::Type{C}) where C<:Colorant eltypec(C)
+@deprecate eltype(::C) where C<:Colorant eltypec(C)
+@deprecate length(::Type{C}) where C<:Colorant lengthc(C)
+@deprecate length(::C) where C<:Colorant lengthc(C)
 
 """
 ColorTypes summary:
@@ -73,7 +80,7 @@ Concrete types:
 - Grayscale types `Gray` and `Gray24` (subtypes of `Color{T,1}`), and
   the corresponding transparent types `AGray`, `GrayA`, and `AGray32`
 
-- Trait functions `eltype`, `length`, `alphacolor`, `coloralpha`,
+- Trait functions `eltypec`, `lengthc`, `alphacolor`, `coloralpha`,
   `color_type`, `base_color_type`, `base_colorant_type`, `ccolor`
 
 - Getters `red`, `green`, `blue`, `alpha`, `gray`, `comp1`, `comp2`, `comp3`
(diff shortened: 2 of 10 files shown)


kimikage:
I'm mostly in favor of it. :+1:
However, I do not feel comfortable with the gradual transition without a determinate goal. I'm against using methods such as `eltypec` and` lengthc`, i.e. using API changes as "notifications".

kimikage:
BTW, when implementing ColorBlendModes, I realized that simple iterators are not very useful other than with RGB colors. For example, the hue is not normalized to [0,1], and in `Lab` and` Luv` spaces, `L` is a non-negative number, but `a`, `b`, `u` and `v` can be both positive and negative. And the alpha is clearly a foreign component as a color component.

timholy:
> However, I do not feel comfortable with the gradual transition without a determinate goal. I'm against using methods such as `eltypec` and `lengthc`, i.e. using API changes as "notifications".

Not entirely sure I understand your meaning. Two stated goals might be (1) to ensure we never treat colorants like containers, or (2) to make colorants behave like numbers wrt the iterator interface. (2) probably requires (1) so it is the more stringent goal (and I'm not sure we want to support `iterate` at all).

However, the operation performed by `eltypec` (extract the numeric type) is certainly useful, independent of what you call it.

kimikage:
I'm sorry I didn't explain it well. I was envisioning a slightly longer term goal, i.e. whether ColorTypes or its add-on provides an API set for treating colorants as containers.

I think that (1) and (2) are appropriate as subgoals. Perhaps the meaning of those subgoals depends on the long-term goal.

> However, the operation performed by `eltypec` (extract the numeric type) is certainly useful, independent of what you call it.

That’s exactly what I was thinking. Since it is useful, I don't think we should give it a name that is reminiscent of `eltype`. Or, if we want it to look like `eltype`, it should definitely be `eltype` (and we will introduce iterator-like or view-like interfaces).

BTW,
```julia
julia> real(Complex{Float32})
Float32
```
So, should `red(RGB{Float32})` and `comp1(RGB{Float32})` return `Float32`? I feel them a bit strange, but they will support the reasonableness of a new method named `component` or `componentof`,

johnnychen94:
My limited understanding couldn't come up with a case that could be done more intuitively and/or concisely compared to the status quo. I'm on sold if there's such one.

kimikage:
As mentioned in https://github.com/JuliaGraphics/ColorTypes.jl/issues/167#issuecomment-593042392, I think that it might be nice that color types can be like tuples 
in the "future". In other words, I think it is a too strong constraint that all components must be of the same type. For example, it makes a trouble with `RGB565`/`RGB332` (cf #114, #115). Some graphics software and image formats encode `Lab`/`YCbCr`-like colors as `{Normed, Fixed, Fixed}`.

Therefore, I think it is a good idea to use this as an opportunity to create an API which assumes that the type of each component can be different, even if we support only homogeneous color types for the time being.

In that sense, I don't think there is a reason to merge this PR immediately.
> or are you saying that by not supporting iterate and the iteration traits, it's essentially OK that we currently use eltype and length? (i.e., that the status quo is actually OK.)

This is not the essence of my idea, but the answer is "yes".:smile:

timholy:
One option to support iteration is a function `channels(c)` that returns a tuple. Probably an important rule would be `typeof(c)(channels(c)...,) === c`.

kimikage:
By the way, ImageCore.jl defines an internal function named `tuplify`.
```julia
tuplify(c::Color1) = (comp1(c),)
tuplify(c::Color3) = (comp1(c), comp2(c), comp3(c))
tuplify(c::Color2) = (comp1(c), alpha(c))
tuplify(c::Color4) = (comp1(c), comp2(c), comp3(c), alpha(c))
```
https://github.com/JuliaImages/ImageCore.jl/blob/3beb2226dce03cdc505cca6d6ce6dc1626b533b1/src/colorchannels.jl#L238-L241

Its implementation is not so good now (cf. #198), and its functionality should be in ColorTypes.jl rather than ImageCore.jl. Therefore, we may define `Tuple(::Colorant)` (cf. https://github.com/JuliaGraphics/ColorVectorSpace.jl/pull/131#discussion_r443177446) independently of the main topic of this issue.
(I have not yet checked on the invalidation, though.)

timholy:
https://discourse.julialang.org/t/rgb-has-a-length-but-no-iterate/84446

johnnychen94:
I'm open to any decision made here and will try to coordinate the downstream JuliaImages ecosystem.

PR #186, opened 2020-04-20
Do not merge: Implement < as well as isless, zero & oneunit for more types

timholy:
#160 "moved" `isless` from ColorVectorSpace to ColorTypes, but didn't likewise move `<`. Because `<` falls back to `isless`, this meant that comparisons like `Gray(0.3) < Gray(NaN)` returned a different answer than `0.3 < NaN`.

This PR also moves code and tests from "traits" to "operations." I should have done these as two separate commits, sorry. The new code & tests are just implementation of `<(::AbstractGray, ::AbstractGray)` and the four tests
```julia
    @test (Gray(0.3) < Gray(NaN)) == (0.3 < NaN)
    @test (Gray(NaN) < Gray(0.3)) == (NaN < 0.3)
    @test isless(Gray(0.3), Gray(NaN)) == isless(0.3, NaN)
    @test isless(Gray(NaN), Gray(0.3)) == isless(NaN, 0.3)
```


src/operations.jl
```diff
@@ -1,3 +1,69 @@
...
+
+zero(::Type{C}) where {C<:Gray} = C(0)
+oneunit(::Type{C}) where {C<:Gray} = C(1)
+
+function Base.one(::Type{C}) where {C<:Gray}
+    Base.depwarn("one($C) will soon switch to returning 1; you might need to switch to `oneunit`", :one)
+    C(1)
+end
+
+Base.broadcastable(x::Colorant) = Ref(x)
```
kimikage, line 62: Do these methods also need to be in "operations.jl"? (I don't think "traits.jl" is the most suitable place, though.)

timholy: `broadcastable` should probably move back to traits.jl. I think the rest may be better here than there, though I'm also open to a different suggestion altogether.

timholy:
This has acquired `zero` and `oneunit` implementations as discussed in https://github.com/JuliaGraphics/ColorTypes.jl/pull/158#discussion_r411331487. @johnnychen94, I also added arithmetic to the error hints.

src/operations.jl
```diff
@@ -1,3 +1,77 @@
...
+    isapprox(gray(a), gray(b); kwargs...)
+_isapprox(TA::Type, TB::Type, a, b; kwargs...) = false
+isapprox(x::Number, y::AbstractGray; kwargs...) =
+    isapprox(x, gray(y); kwargs...)
+isapprox(x::AbstractGray, y::Number; kwargs...) =
+    isapprox(y, x; kwargs...)
+
+zero(::Type{C}) where {C<:ColorN{1}} = C(0)
+zero(::Type{Gray24})                 = Gray24(zero(N0f8))
+zero(::Type{C}) where {C<:ColorN{3}} = C(0, 0, 0)
```
johnnychen94, line 55: This entails that `zeros(RGB, 4, 4)` works while `ones(RGB, 4, 4)` doesn't.

src/operations.jl
```diff
@@ -1,3 +1,77 @@
...
+# It's not entirely clear what `oneunit` means for most Color3s, but for AbstractRGB it's OK
+oneunit(::Type{C}) where {C<:AbstractRGB}     = C(1, 1, 1)
+oneunit(::Type{C}) where {C<:TransparentGray} = C(oneunit(color_type(C)), 1)
+oneunit(::Type{C}) where {C<:TransparentRGB}  = C(oneunit(color_type(C)), 1)
+oneunit(c::Colorant)                          = oneunit(typeof(c))
+
+function Base.one(::Type{C}) where {C<:Gray}
+    Base.depwarn("one($C) will soon switch to returning 1; you might need to switch to `oneunit`", :one)
+    C(1)
+end
```
johnnychen94, line 70: One more method?
```julia
one(c::Gray) = one(typeof(c))
```

src/operations.jl
timholy: I know. I dislike `ones` :smile:.

src/operations.jl
timholy: Oh shoot I forgot about #175. Sorry!

WRT 1, since this package doesn't support multiplication, what should be returned for the multiplicative identity?

johnnychen94: Hmmm, are you indicating that `one(::Type{<:Gray}) = C(1)` should be moved to ColorVectorSpace without depwarn?

johnnychen94: actually, why don't we add `white`/`whites`/`black`/`blacks` as the direct replacements of `one`/`ones`/`zero`/zeros`?

kimikage: For example, what should `white` or `black` return in `HSL`?
After all, if they are obvious, abstraction is not really helpful.

src/operations.jl
```diff
@@ -1,3 +1,77 @@
...
+_isapprox(TA::Type, TB::Type, a, b; kwargs...) = false
+isapprox(x::Number, y::AbstractGray; kwargs...) =
+    isapprox(x, gray(y); kwargs...)
+isapprox(x::AbstractGray, y::Number; kwargs...) =
+    isapprox(y, x; kwargs...)
+
+zero(::Type{C}) where {C<:ColorN{1}} = C(0)
+zero(::Type{Gray24})                 = Gray24(zero(N0f8))
+zero(::Type{C}) where {C<:ColorN{3}} = C(0, 0, 0)
+# It's not clear whether zero(TransparentColor) should use alpha = 0 or alpha = 1
```
kimikage, line 56: Since general alpha compositing is not commutative, I can hardly imagine additive groups in which `alpha = 1` is valid. `a1 * a2` or `min(a1, a2)`?

src/operations.jl
johnnychen94: `HSL(gamutmin(HSL)...)` and `HSL(gamutmax(HSL)...)` looks good to me.

The point of introducing these two convenience function is, if people ask you what's the difference between `one`/`oneunit` and why `zeros` works while `ones` doesn't, you could save yourself some efforts and tell them to use `blacks` and `whites`. Not everyone is color expert and people from other languages do miss `ones` when they actually want `whites`.

kimikage: I consider `zeros` and `ones` to be array initializers, so I don't think anything requiring additional documentation is appropriate for the "identity". We implicitly or informally expect identities to be unique.
Therefore, I think there may be `black`/`white` which returns "an example" of black / white, but `blacks`/`whites` is never better than `zeros`/`ones`.

BTW, unfortunately, we already have the `red`/`green`/`blue`. The `black`/`white` will introduce inconsistencies.:ghost:

src/operations.jl
timholy: Yeah, first thing I did this morning was re-read the docstring for `zero`, and indeed it's clear it should be `alpha = 0`. In an additive accumulator, nothing else would make sense.

Interesting, though, that it's phrased as the additive identity. (Totally makes sense, of course.) From a strictly logical standpoint, having all of these defined only in ColorVectorSpace seems like it might be most consistent. It might prove to be a pain in practice, though.

src/error_hints.jl
```diff
@@ -1,12 +1,21 @@
+MathTypes{T,C} = Union{AbstractRGB{T},TransparentRGB{C,T},AbstractGray{T},TransparentGray{C,T}}
+
 # provided by https://github.com/JuliaLang/julia/pull/35094
 function register_hints()
```
johnnychen94, line 4: This is likely to be an experimental feature: https://github.com/JuliaLang/julia/issues/35671

kimikage: https://github.com/JuliaLang/julia/pull/35679 was merged. So, I think it should be fixed. However, I don't know whether we can continue to maintain such experimental features. The nightly build is always experimental, though.
 cc @johnnychen94, @timholy

timholy: #193

timholy:
This has been cleaned up and is being proposed for merging. In implementing https://github.com/JuliaGraphics/ColorVectorSpace.jl/issues/126 I've decided not to support `*` for RGB values (`⋅`, `⊙`, and `⊗` are supported instead), and so for consistency I think it's best to follow through on our deprecation warning for `one(::Gray) = 1`. Assuming we merge #177 (which is breaking), indeed I think it might be time to remove the deprecation warning? It's been in place since 2017, which seems plenty long.

kimikage:
I'm checking just to be sure, but does removing the deprecation warning mean the decision to define `one = oneunit` for grays?

If we don't deprecate the `one` itself, I think we should generalize it for `AbstractGray`(`ColorN{1}`) and support the instances. (cf. https://github.com/JuliaGraphics/ColorTypes.jl/pull/158#discussion_r411279179)

timholy:
`one(::Gray)` already warns that it will switch from returning `Gray(1)` to `1`. I'm proposing we follow through with the depwarn and make the change.

timholy:
I should clarify, though, that this decision isn't obvious. The other way to go is to have it return `Gray(1)`. It's viable because we can and perhaps should define `*(::Gray, ::Gray)` (in ColorVectorSpace), whereas I'm pretty sure we want to continue to not define `*(::RGB, ::RGB)`. So `one(::RGB)` *definitely* has to return 1; the question is, what should `one(::Gray)` return? I'm personally included to make it be 1 for consistency, even though `*` will be supported for `Gray` but not `RGB`.

johnnychen94:
I still feel `one(Gray(1))` should return `Gray(1)`. The inconsistency between `one(RGB(1)) == 1` and `one(Gray(1)) == Gray(1)` doesn't cause much trouble in practice since `Gray(1) == 1` and multiplication is poisoning. But `ones(4, 4)` and `zeros(4, 4)` returns both `Array{Gray, 2}` is more important in practice.

kimikage:
I have no strong policy on the definition of `one`.:sweat_smile:
FYI, even if we define `one(::Type{<:Colorant}) = 1`, the following works. (cf. https://github.com/JuliaGraphics/ColorTypes.jl/pull/175#discussion_r386119901)
```julia
julia> ones(RGB,2,2)
2×2 Array{RGB,2} with eltype RGB:
 RGB{N0f8}(1.0,1.0,1.0)  RGB{N0f8}(1.0,1.0,1.0)
 RGB{N0f8}(1.0,1.0,1.0)  RGB{N0f8}(1.0,1.0,1.0)
```

kimikage:
BTW, are the specialized `zero` and `one` for `Gray24` required?

As for `zero`, I think they can be simplified and generalized as follows.
```julia
zero(::Type{C}) where {C<:ColorantN{1}} = C(0)
zero(::Type{C}) where {C<:ColorantN{2}} = C(0, 0)
zero(::Type{C}) where {C<:ColorantN{3}} = C(0, 0, 0)
zero(::Type{C}) where {C<:ColorantN{4}} = C(0, 0, 0, 0)
zero(::Type{C}) where {C<:ColorantN{5}} = C(0, 0, 0, 0, 0) # issue #179
```
Alternatively, it might be nice to add no-argument constructors for the opaque `Color` types. Currently,
```julia
julia> RGB()
RGB{N0f8}(0.0,0.0,0.0)

julia> RGB{Float32}()
ERROR: MethodError: no method matching RGB{Float32}()
```

johnnychen94:
I didn't know that `ones(RGB,2,2)` works, and it seems not to work now, either. Which version do you use?

```julia
julia> using ColorVectorSpace, ImageCore

julia> ones(RGB,2,2)
ERROR: MethodError: no method matching one(::Type{RGB})
Closest candidates are:
  one(::Type{Missing}) at missing.jl:103
  one(::BitArray{2}) at bitarray.jl:422
  one(::Missing) at missing.jl:100
  ...
Stacktrace:
 [1] ones(::Type{RGB}, ::Tuple{Int64,Int64}) at ./array.jl:505
 [2] ones(::Type{RGB}, ::Int64, ::Int64) at ./array.jl:500
 [3] top-level scope at REPL[3]:1

(@v1.4) pkg> st ImageCore ColorTypes Colors ColorVectorSpace
Status `~/.julia/environments/v1.4/Project.toml`
  [3da002f7] ColorTypes v0.10.4
  [c3611d14] ColorVectorSpace v0.8.5
  [5ae59095] Colors v0.12.2
  [a09fc81d] ImageCore v0.8.14
```

timholy:
Just to throw out another thought that's completely the opposite of where I started, we now have
```julia
julia> RGB(1)
RGB{N0f8}(1.0,1.0,1.0)
```
so if we make `*` mean elementwise multiplication then `RGB(1)` is a multiplicative identity.

johnnychen94:
> if we make * mean elementwise multiplication then RGB(1) is a multiplicative identity.

This sounds the most promising solution to me.

kimikage:
> I didn't know that ones(RGB,2,2) works, and it seems not to work now, either. Which version do you use? Or did I miss something?

@johnnychen94,

> if we define `one(::Type{<:Colorant}) = 1`

:smile:

Making `*` element-wise partially messes up the discussions in https://github.com/JuliaGraphics/ColorVectorSpace.jl/issues/126 and https://github.com/JuliaLang/julia/pull/35150. :confused:

johnnychen94:
I see, but that's not a good practice in performance:

```julia
julia> @btime ones(RGB,512,512);
  1.119 ms (262146 allocations: 6.00 MiB)

julia> @btime fill(RGB{N0f8}(1.0, 1.0, 1.0), 512, 512);
  141.963 μs (2 allocations: 768.08 KiB)
```

kimikage:
See https://github.com/JuliaGraphics/ColorTypes.jl/pull/175#discussion_r386119901 again.
```julia
julia> @btime ones(RGB{N0f8},512,512);
  17.400 μs (2 allocations: 768.08 KiB)
```

kimikage:
In any case, this PR should not depend on the definition of `one` or color multiplication.
I think it is better to discuss and decide the system of color arithmetic, in ColorVectorSpace. I also want to organize the release schedule (the order, not the dates) of these color/image related packages.

timholy:
To assist in such deliberations, I decided to submit my PR even though it has a lot of unmerged dependencies: https://github.com/JuliaGraphics/ColorVectorSpace.jl/pull/131

timholy:
Are we to the point of making breaking changes here again? Or do you still have a queue you want to merge first?

kimikage:
> Are we to the point of making breaking changes here again?

I'm not sure, since I released v0.10.7 and v0.10.8 even after saying to proceed to the v0.11 series.:sweat_smile:
However, I believe I am standing at v0.11.0-DEV now. :smiley:

timholy:
OK, starting Wednesday I'll get back to this and trying to finish off the new version of ColorVectorSpace.

kimikage:
By the way, we can slightly generalize `==`, now (cf. #198).
```julia
_is_same_colorspace(a, b) = base_colorant_type(a) === base_colorant_type(b)
_is_same_colorspace(a::AbstractRGB, b::AbstractRGB) = true
# ...snip...
==(a::ColorantN{3}, b::ColorantN{3}) = _is_same_colorspace(a, b) &&
    (comp1(a) == comp1(b)) & (comp2(a) == comp2(b)) & (comp3(a) == comp3(b))
# ...snip...
```
Shall I submit a separate PR?

kimikage:
> I'm not sure, since I released v0.10.7 and v0.10.8 even after saying to proceed to the v0.11 series.:sweat_smile:
> However, I believe I am standing at v0.11.0-DEV now. :smiley:

Umm...the devil in me is starting to whisper that the `<` is a v0.10 material.

On the other hand, I don't think any non-trivial or non-editorial changes of error hints (e.g. the `one`'s one) are the patch level, even if they don't break the behavior.

timholy:
This will be superseded by a set of independent smaller PRs. I'll leave this open until they're all merged.

kimikage:
PRs separated from this PR:
- [x] #213
- [x] #238
- [x] #249

The discussion on `one` is continued in issue #235 and PR #243.

Issue #179, opened 2020-04-08
Support for 4-component color types


kimikage:
There has been a request for `CMYK` support for some time (cf. https://github.com/JuliaGraphics/Colors.jl/issues/264).
IMO, Colors.jl, Images.jl and so on still need some time to properly handle the `CMYK` colors, so we should not add `CMYK` to ColorTypes *for now*.

However, it is good that users can easily define and use 4-component (or 2-component) colors.

It is good to specialize the 3-component or 1-component colors in terms of performance, but at present, they are also specialized in terms of functionalities, i.e. they lack generality.
For example, 4-component colors cannot be shown properly.
https://github.com/JuliaGraphics/ColorTypes.jl/blob/8afb1507c4a3e1f77dd7131ea5771963be5e6cc3/src/show.jl#L9-L12

kimikage:
Why doesn't `throw_colorerror`(and its caller `checkval`) use the color types directly?
https://github.com/JuliaGraphics/ColorTypes.jl/blob/4bb8aa9c918c376a14c2717c4e9edef67224f94c/src/types.jl#L683-L695

kimikage:
Anyway, in order to handle types like `ACMYK{N0f8}`, we need to add a 5-arg `checkval`.
Since Julia doesn't seem to do "horizontal" SIMD optimizations, I think there is room for optimizations.
https://github.com/JuliaGraphics/ColorTypes.jl/blob/4bb8aa9c918c376a14c2717c4e9edef67224f94c/src/types.jl#L644-L652

timholy:
> Why doesn't throw_colorerror(and its caller checkval) use the color types directly?

Clearly it should!

johnnychen94:
> IMO, Colors.jl, Images.jl and so on still need some time to properly handle the CMYK colors, so we should not add CMYK to ColorTypes for now.

I don't quite get the reason behind, could you please give some further explanation on this?

kimikage:
In the first place, Colors.jl does not have "practical" conversion methods between CMYK space and other color spaces. I think that CMYK conversion which is not based on color profiles is just a toy. :smile:

> Do you mean adding CMYK to ColorTypes doesn't make it generally a valid input to functions in Colors and Images because it has 4 components?

At least I don't think that has been fully tested. It is quite possible that there are codes other than the above which interpret the 4-component colors as `RGBA` or `Transparent3` colors. Moreover, `TransparentCMYK` is a 5-component color.

[...]

PR #189, opened 2020-04-21
Fix `show` for colors with uncommon components and `AGray32`

kimikage:
This generalizes and simplifies the implementation of `show`.(cf. https://github.com/JuliaGraphics/ColorTypes.jl/issues/179#issuecomment-617117028, https://github.com/JuliaGraphics/ColorTypes.jl/issues/179#issuecomment-617131514)
This also adds `comp4` and `comp5`, and enables `color` for all colors with 5 or fewer components.

For `Transparent3` colors, `comp4` returns the `alpha`s. Therefore, after merging this PR, `mapc`, `reducec` and `mapreducec` can be defined more consistently (in lexical terms).

### DIFF
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -56,9 +56,11 @@ hue(c::Union{C,AlphaColor{C},ColorAlpha{C}}) where {C<:Union{Lab,DIN99,DIN99o,DI
 hue(c::Union{C,AlphaColor{C},ColorAlpha{C}}) where {C<:Luv} =
     (h = atand(c.v, c.u); h < 0 ? h + 360 : h)
 
+# fallbacks for compN
+_comp(::Val{N}, c::Colorant) where N = getfield(c, N)
+_comp(::Val{N}, c::AlphaColor) where N = getfield(c, N + 1)
+_comp(::Val{N}, c::AlphaColorN{N}) where N = alpha(c)
 
-# Extract the first, second, and third arguments as you'd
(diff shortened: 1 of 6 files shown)


src/traits.jl
```diff
@@ -56,9 +56,11 @@ hue(c::Union{C,AlphaColor{C},ColorAlpha{C}}) where {C<:Union{Lab,DIN99,DIN99o,DI
 hue(c::Union{C,AlphaColor{C},ColorAlpha{C}}) where {C<:Luv} =
     (h = atand(c.v, c.u); h < 0 ? h + 360 : h)
 
+# fallbacks for compN
+_comp(::Val{N}, c::Colorant) where N = getfield(c, N)
+_comp(::Val{N}, c::AlphaColor) where N = getfield(c, N + 1)
+_comp(::Val{N}, c::AlphaColorN{N}) where N = alpha(c)
```
kimikage, line 62: We may need a parametric interface like this, not for fallback. In fact, The "ugly" tuple `comp_n = (comp1, comp2, comp3, comp4, comp5)` was used in `_show_components()`.

However, when users overload the getters, I think the nonparametric methods (e.g. `comp1`) are simpler. (`comp1` might already have been overloaded somewhere. :ghost:) So I think `comp1` etc. should be left as independent methods, not as aliases.

kimikage:
I would like to split this into several PRs.

PR #193, opened 2020-05-04
Fix on nightly

### LINKED ISSUE #192: register_error_hint not defined.
On Julia master 1.5.0-DEV.823, [this line](https://github.com/JuliaGraphics/ColorTypes.jl/blob/23cae4e94a4189fbd41ed8688a2280c691d6d8e7/src/error_hints.jl#L4) should probably read `Base.Experimental.register_error_hint`.

timholy:
Fixes #192. The strategy used here has the advantage of being safer against future changes.

### DIFF
--- a/src/error_hints.jl
+++ b/src/error_hints.jl
@@ -1,7 +1,7 @@
 # provided by https://github.com/JuliaLang/julia/pull/35094
 function register_hints()
-    if VERSION >= v"1.5.0-DEV.491"
-        register_error_hint(MethodError) do io, exc, argtypes, kwargs
+    if isdefined(Base, :Experimental) && isdefined(Base.Experimental, :register_error_hint)
+        Base.Experimental.register_error_hint(MethodError) do io, exc, argtypes, kwargs
             if exc.f in (zero, one) && argtypes[1] <: Union{Type{<:AbstractRGB}, AbstractRGB}
                 print(io, "\nYou may need to `using ColorVectorSpace`.")
             end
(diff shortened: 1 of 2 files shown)


timholy:
Also includes minor invalidation-related changes.

kimikage:
> Also includes minor invalidation-related changes.

I think it's better to split the PR. The changes themselves are trivial, but I don't think the reason for them (i.e. invalidation) is trivial.

timholy:
They're two separate commits, and I wasn't planning on squashing when merging. But I can split in two PRs if you prefer. Will do tomorrow.

kimikage:
If the subject of this PR (or the merge commit) is clear, the splitting is optional.
In any case, this PR should be modified.
```
ERROR: LoadError: InitError: UndefVarError: Experimental not defined
```

src/error_hints.jl
```diff
@@ -1,7 +1,7 @@
 # provided by https://github.com/JuliaLang/julia/pull/35094
 function register_hints()
-    if VERSION >= v"1.5.0-DEV.491"
-        register_error_hint(MethodError) do io, exc, argtypes, kwargs
+    if isdefined(Base.Experimental, :register_error_hint)
```
johnnychen94, line 3: `Base.Experimental` module doesn't exist until julia-1.3, so we might still want to keep the version check here.

timholy: Thanks @johnnychen94. I decided to check for the existence of `Experimental` instead. The thing I like about this approach is it's essentially a direct test of "will the following code work?" Version checks are a bit more indirect, and if you want to understand the meaning of the particular version you might have to do some git archeology (though a comment with hyperlink can fix that). Of course, they're extremely useful, especially in cases where there isn't a good alternative way to check.

timholy:
Ooh, charming segfault on 1.4! It's due to the invalidations commit, I will indeed separate that out.

PR #194, opened 2020-05-05
Reduce invalidations of other methods

timholy:
Ref https://github.com/JuliaLang/julia/pull/35733. This will have very little impact on its own, but it shouldn't hurt anything and it prepares the way for future gains.

Based on #193 and local tests I expect this to crash on 1.4, so for now this is mostly posted as a reference for a bug report.

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -70,7 +70,8 @@ end
 # no-op and element-type conversions, plus conversion to and from transparency
 # Colorimetry conversions are in Colors.jl
 convert(::Type{C}, c::C) where {C<:Colorant} = c
-convert(::Type{C}, c) where {C<:Colorant} = cconvert(ccolor(C, typeof(c)), c)
+convert(::Type{C}, c::Colorant) where {C<:Colorant} = cconvert(ccolor(C, typeof(c)), c)
+convert(::Type{C}, c::Number) where {C<:Colorant} = cconvert(ccolor(C, typeof(c)), c)
 cconvert(::Type{C}, c::C) where {C} = c
 cconvert(::Type{C}, c) where {C} = _convert(C, base_color_type(C), base_color_type(c), c)
 
--- a/src/ColorTypes.jl
+++ b/src/ColorTypes.jl
@@ -81,7 +81,8 @@ Concrete types:
 Use `?` to get more information about specific types or functions.
 """ ColorTypes
 
-if VERSION >= v"1.1" # work around https://github.com/JuliaLang/julia/issues/34121
+if VERSION >= v"1.4.2" # work around https://github.com/JuliaLang/julia/issues/34121
+    # ...and also requires https://github.com/JuliaLang/julia/pull/35378
     include("precompile.jl")
     _precompile_()
 end


timholy:
Also segfaults a binary of Julia 1.2. But not 1.0.

timholy:
Fixed by https://github.com/JuliaLang/julia/pull/35378

kimikage:
Do we have to wait for the release of v1.4.2 or v1.5.0?
I'm not sure which one is critical, the precompilation or the invalidation.

timholy:
Crashes bad. My guess is we'll have make these methods conditional on `VERSION >= v"1.5"`, because I'm pretty sure there won't be backports to 1.2.

kimikage:
Is the precompilation effective only for top level, i.e. public methods?

src/conversions.jl
```diff
@@ -70,7 +70,8 @@ end
 # no-op and element-type conversions, plus conversion to and from transparency
 # Colorimetry conversions are in Colors.jl
 convert(::Type{C}, c::C) where {C<:Colorant} = c
-convert(::Type{C}, c) where {C<:Colorant} = cconvert(ccolor(C, typeof(c)), c)
+convert(::Type{C}, c::Colorant) where {C<:Colorant} = cconvert(ccolor(C, typeof(c)), c)
+convert(::Type{C}, c::Number) where {C<:Colorant} = cconvert(ccolor(C, typeof(c)), c)
 cconvert(::Type{C}, c::C) where {C} = c
```
kimikage, line 75: BTW, I think this method is dead, based on the current `ccolor` implementation. (`convert(::Type{C}, c::C) where {C<:Colorant}` has priority.) The re-examination of the conversion methods may be necessary.

timholy:
> Is the precompilation effective only for top level, i.e. public methods?

The issue here is that defining this method invalidates a whole bunch of other methods that Julia has already compiled. If you have a <1day old master of Julia, you can try the experiment [here](https://gist.github.com/timholy/8e664ae78fc504049e27256453bfccd3). The output is a list of methods that are no longer valid and must be recompiled. As you can see, it's nearly 3000 methods.

kimikage:
I mean, does not it make sense to precompile `cconvert` instead of `convert` in this package?
I still haven't studied enough on the Julia internals...

kimikage:
I'm looking for a workaround for the "crash" rather than the invalidation. 
I don't want different method definitions depending on the version.

timholy:
There may be a good fix, see https://github.com/JuliaLang/julia/pull/35768#issuecomment-624846600

kimikage:
> Is the precompilation effective only for top level, i.e. public methods?

> does not it make sense to precompile `cconvert` instead of `convert` in this package?

After all, what is the conclusion of these? To my understanding, precompiled `cconvert` seems to be as effective (or, not so effective) as precompiled `convert`.

timholy:
I've now restricted precompilation to v1.4.2 or higher. That should be a pretty safe fix.

src/conversions.jl
timholy: I tried deleting them but without other changes they are still necessary. Any other cleanup can happen in a separate PR.

PR #196, opened 2020-05-18
Move type-pirating `convert` methods from ImageCore

timholy:
This removes 80 invalidations, ref https://github.com/JuliaLang/julia/pull/35922  (EDIT, oops, wrote 35915 first)

This will be breaking so I've bumped the version in anticipation. We might want to merge this only when we're ready to release the next breaking version.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ColorTypes"
 uuid = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
-version = "0.10.3"
+version = "0.11.0"
 
 [deps]
 FixedPointNumbers = "53c48c17-4a7d-5ca2-90c5-79b7896eea93"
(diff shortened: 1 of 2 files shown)


kimikage:
Did you mean https://github.com/JuliaLang/julia/issues/35922?

JeffBezanson:
Is it possible these could be removed? It seems like the only thing they do is add a call to `ccolor`. I don't know how important that is, but it's some pretty heavy piracy (since AFAICT it will not always convert to exactly the requested type).

timholy:
Yes, I'd *love* to remove these, and it seems like it could be possible in theory. What `ccolor` lets you do is specify the "outer" type container without specifying the inner one: `convert(RGB, c::HSV{Float32})` rather than needing to say `convert(RGB{Float32}, c)`. It picks the precision for you. That turns out to be pretty important to usability in writing generic code, but writing `ccolor` was painful.

Now that I think about it, I believe these all date from an era long before we had dot-syntax or had resolved the issues with eltypes of empty containers. These days, even this works:
```julia
julia> a = rand(HSV{Float32}, 0, 0)
0×0 Array{HSV{Float32},2} with eltype HSV{Float32}

julia> RGB.(a)
0×0 Array{RGB{Float32},2} with eltype RGB{Float32}
```
so perhaps we could just get rid of them after a suitable deprecation period.

PR #197, opened 2020-06-23
[RFC] Redesign color constructors

### LINKED ISSUE #156: Constructing transparent colors with FixedPoint arguments

kimikage:
This eliminates the macro-based constructor generation. This helps add user-defined types (cf. #179).
This also allows grays to be used in the constructor argument in more cases (cf. https://github.com/JuliaGraphics/ColorTypes.jl/pull/177#discussion_r412868245).

Fixes #156

### DIFF
--- a/README.md
+++ b/README.md
@@ -423,4 +423,4 @@ In most cases, adding a new color space is quite straightforward:
 In special cases, there may be other considerations:
 - For `AbstractRGB`/`AbstractGray` types, `0` means "black" and `1` means
   "saturated."
-- If your type has extra fields, check the "Generated code" section of `types.jl` carefully. You may need to define a `colorfields` function and/or call `@make_constructors` or `@make_alpha` manually.
+- If your type has extra fields, check the "Generated code" section of `types.jl` carefully. You may need to define a `colorfields` function and/or call `@make_alpha` manually.
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -76,7 +76,6 @@ cconvert(::Type{C}, c) where {C} = _convert(C, base_color_type(C), base_color_ty
 
 convert(::Type{C}, c::Color, alpha) where {C<:TransparentColor} = cconvert(ccolor(C, typeof(c)), c, alpha)
 cconvert(::Type{C}, c::Color, alpha) where {C<:TransparentColor} =_convert(C, base_color_type(C), base_color_type(c), c, alpha)
-cconvert(::Type{ARGB32}, c::AbstractRGB, alpha) = ARGB32(c, alpha) # optimization for speed
 
 # Fallback definitions that print nice error messages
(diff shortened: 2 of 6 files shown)


timholy:
Took a glance (didn't review since it's marked Draft), this looks quite promising!

kimikage:
I made the ambiguity resolution simpler than when this PR was originally submitted. Now it works without explicitly writing constructors. This means that there is no ambiguity with the default constructors.

johnnychen94:
Simplifying the macro looks promising to me. Just a simple loading time test on this that might worth noting. (No objection on this PR)

```julia
julia> versioninfo()
Julia Version 1.6.0-rc1
Commit a58bdd9010 (2021-02-06 15:49 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin18.7.0)
  CPU: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, skylake)

julia> @time using ColorTypes
  # master:  0.266565 seconds (506.97 k allocations: 37.554 MiB, 5.03% gc time, 18.80% compilation time)
  # PR 0.308603 seconds (646.04 k allocations: 50.311 MiB, 4.10% gc time, 13.67% compilation time)
```

kimikage:
I'm not going to make any performance related tweaks in this PR unless there is a very large regression.
This PR is not an end goal, but a means to solve a problem. Subsequent changes may improve the performance, or conversely, may significantly degrade it.:smile:

kimikage:
This is a drastic change and I have not been able to fully understand the impact of it.
Although I have changed the commit of this PR many times, the essential points have not changed from the original commit. And I have been experimenting with this change for the past year, and have not encountered any fatal problems.

Well, I've been struggling with the precompilation problems in Colors.jl for over a year now. I don't exactly know the impact of this change, but at least I know this won't have a fatal negative impact.

So, I plan to merge this in around this weekend.

cc: @johnnychen94

johnnychen94:
I believe we'll live well with it.

kimikage:
I'll merge this, but it won't be released for a while yet.
The promotion rules for constructor arguments will be changed before the release of v0.12.

Issue #190, opened 2020-04-22
[RFC] Adding component iterator


kimikage:
This is a PoC to clarify the discussion in #184, and is the modernized (Julia v1) version of #7. See also #119

```julia
using ColorTypes
import ColorTypes.ColorantN

struct ComponentIterator{C<:Colorant}
    c::C
end

function Base.getproperty(c::C, sym::Symbol) where C <: Colorant
   if sym === :components
      return ComponentIterator{C}(c)
   else # fallback to getfield
      return getfield(c, sym)
   end
end

Base.eltype(itr::ComponentIterator{C}) where {T, C <: Colorant{T}} = T
Base.length(itr::ComponentIterator{C}) where {N, C <: ColorantN{N}} = N

function Base.iterate(itr::ComponentIterator{C}, state::Int=0) where {N, C <: ColorantN{N}}
    state >= N && return nothing
    state += 1
    (itr[state], state)
end

function Base.getindex(itr::ComponentIterator{C}, i::Integer) where {N, C <: ColorantN{N}}
    comp_n = (comp1, comp2, comp3) # FIXME
    (comp_n[i])(itr.c)
end
```

```julia
julia> rgb_f32 = RGB{Float32}(1.0, 0.5, 0.0)
RGB{Float32}(1.0f0,0.5f0,0.0f0)

julia> eltype(rgb_f32.components)
Float32

julia> length(rgb_f32.components)
3

julia> for c in rgb_f32.components
           @show c
       end
c = 1.0f0
c = 0.5f0
c = 0.0f0

julia> rgb_f32.components[2]
0.5f0

julia> RGB((RGB(1.0, 0.5, 0.0).components .+ RGB(0.0, 0.1, 0.2).components)...)
RGB{Float64}(1.0,0.6,0.2)
```

The above has some problems in practical use, but IMO, it is better than `eltypec` and `lengthc`.

kimikage:
It maight be a good idea to create a new package ColorComponentIterators.jl, as ColorVectorSpace.jl is a separate package.
In fact, suc a simple iterator, `mapc`, and so on are not very useful (cf. https://github.com/JuliaGraphics/ColorTypes.jl/pull/184#issuecomment-616175350). More complex iterators may be needed.

timholy:
With constant propagation this might be more useful than previously, and hiding it behind a `c.components` seems like a pretty clever design. Thanks for submitting this, I will have to think through the implications.

[...]

PR #198, opened 2020-06-23
Add component accessor `comp4`/`comp5`

kimikage:
This is a modified version of part of PR #189, but this does not change the `show()`.

In the future, I want to be able to get an iterator or tuple of the components (cf. #190, https://github.com/JuliaGraphics/ColorVectorSpace.jl/pull/131#discussion_r443177446).

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,8 +402,8 @@ greater detail); just type `?ccolor` at the REPL.
 - `alpha` extracts the alpha channel from any `Color` object
   (returning 1 if there is no alpha channel)
 
-- `comp1`, `comp2`, and `comp3` extract color components in the order
-  expected by the constructor
+- `comp1`, `comp2`, `comp3`, `comp4` and `comp5` extract color components in the
+  order expected by the constructor
 
 ### Functions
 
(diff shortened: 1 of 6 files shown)


kimikage:
> Would be better to provide an informative and friendly error message:
> 
> ```julia
> julia> comp5(RGB(1, 1, 1))
> ERROR: BoundsError: attempt to access RGB{FixedPointNumbers.Normed{UInt8,8}}
>   at index [5]
> Stacktrace:
>  [1] _comp(::Val{5}, ::RGB{FixedPointNumbers.Normed{UInt8,8}}) at /Users/jc/Documents/Julia/ColorTypes.jl/src/traits.jl:60
>  [2] comp5(::RGB{FixedPointNumbers.Normed{UInt8,8}}) at /Users/jc/Documents/Julia/ColorTypes.jl/src/traits.jl:94
>  [3] top-level scope at REPL[2]:1
> ```

_Originally posted by @johnnychen94 in https://github.com/JuliaGraphics/ColorTypes.jl/pull/199#issuecomment-648564683_

kimikage:
I added the "ugly" error handling for `compN`.
```julia
julia> comp4(XRGB(1,0,0))
ERROR: ArgumentError: attempt to access 3-component color XRGB{Normed{UInt8,8}} with `comp4`
```
This changes the error type from `BoundsError` to `ArgumentError`, but I don't think this is a breaking change because there was no test for `compN`.

kimikage:
Oops, when running `rebase`, I applied the one before the change of the error message.:scream:

Let's redo it in PR #199.

Issue #191, opened 2020-04-26
Display of gray arrays and `Gray{Bool}`


kimikage:
The current implementation:
```julia
julia> Gray{Float32}[1 0; 0.2 0.5]
2×2 Array{Gray{Float32},2} with eltype Gray{Float32}:
 Gray{Float32}(1.0)  Gray{Float32}(0.0)
 Gray{Float32}(0.2)  Gray{Float32}(0.5)

julia> Gray24[1 0; 0.2 0.5]
2×2 Array{Gray24,2} with eltype Gray24:
 Gray24(1.0)  Gray24(0.0)
 Gray24(0.2)  Gray24(0.502)

julia> Gray{Bool}(1)
Gray{Bool}(true)

julia> Gray{Bool}[1 0; 0 1]
2×2 Array{Gray{Bool},2} with eltype Gray{Bool}:
 Gray{Bool}(true)   Gray{Bool}(false)
 Gray{Bool}(false)  Gray{Bool}(true)
```
The implementations of `Real` types:
```julia
julia> Float32[1 0; 0.2 0.5]
2×2 Array{Float32,2}:
 1.0  0.0
 0.2  0.5

julia> Bool(1)
true

julia> Bool[1 0; 0 1] # Julia v1.0.5
2×2 Array{Bool,2}:
  true  false
 false   true

julia> Bool[1 0; 0 1] # Julia v1.4.1
2×2 Array{Bool,2}:
 1  0
 0  1
```
As shown above, an array of `Gray` can be constructed from an array of `Real` numbers, because we have the `convert` method:
https://github.com/JuliaGraphics/ColorTypes.jl/blob/23cae4e94a4189fbd41ed8688a2280c691d6d8e7/src/conversions.jl#L108

Therefore, we can simplify the outputs to:
```julia
julia> Gray{Float32}[1 0; 0.2 0.5]
2×2 Array{Gray{Float32},2}:
 1.0  0.0
 0.2  0.5

julia> Gray24[1 0; 0.2 0.5]
2×2 Array{Gray24,2}:
 1.0  0.0
 0.2  0.5

julia> Gray{Bool}(1)
Gray{Bool}(1)

julia> Gray{Bool}[1 0; 0 1]
2×2 Array{Gray{Bool},2}:
 1  0
 0  1
```
("with eltype *" is optional.)

[...]

PR #199, opened 2020-06-24
Generalize `show` for colors with uncommon components

kimikage:
This is the successor to #198 and the rest of PR #189, but the bug fix for `show(::AGray32)` is removed.

### DIFF
--- a/src/show.jl
+++ b/src/show.jl
@@ -1,49 +1,32 @@
-show(io::IO, c::Colorant) = get(io, :compact, false) ? _showcompact(io, c) : _show(io, c)
-show(io::IO, c::ColorantNormed) = show_normed(io, c)
 
-# Nonparametric types
-show_normed(io::IO, c::Gray24) = print(io, "Gray24(", gray(c), ')')
-show_normed(io::IO, c::RGB24)  = print(io, "RGB24(", red(c), ',', green(c), ',', blue(c), ')')
-show_normed(io::IO, c::ARGB32) = print(io, "ARGB32(", red(c), ',', green(c), ',', blue(c), ',', alpha(c), ')')
+function show(io::IO, c::Colorant)
+    show_colorant_string_with_eltype(io, typeof(c))
+    _show_components(io, c)
+end
 
-# FIXME: handle `Color` and `TransparentColor` correctly
-#       (e.g. `Color{T,4}` such as CMYK is different from `Transparent3`).
-for N = 1:4
-    component = N >= 3 ? (:comp1, :comp2, :comp3, :alpha) : (:comp1, :alpha)
-    printargs = Array{Any}(undef, 2, N)
+# Special handling for Normed types: don't print the giant type name
+function show(io::IO, c::ColorantNormed)
+    show_colorant_string_with_eltype(io, typeof(c))
+    if typeof(c) <: Union{RGB24, ARGB32, Gray24} # isempty(typeof(c).parameters) # Nonparametric types
+        _show_components(io, c) # with trailing "N0f8" unless :compat=>true
+    else
+        _show_components(IOContext(io, :compact=>true), c)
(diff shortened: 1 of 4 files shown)


johnnychen94:
Would be better to provide an informative and friendly error message:

```julia
julia> comp5(RGB(1, 1, 1))
ERROR: BoundsError: attempt to access RGB{FixedPointNumbers.Normed{UInt8,8}}
  at index [5]
Stacktrace:
 [1] _comp(::Val{5}, ::RGB{FixedPointNumbers.Normed{UInt8,8}}) at /Users/jc/Documents/Julia/ColorTypes.jl/src/traits.jl:60
 [2] comp5(::RGB{FixedPointNumbers.Normed{UInt8,8}}) at /Users/jc/Documents/Julia/ColorTypes.jl/src/traits.jl:94
 [3] top-level scope at REPL[2]:1
```

kimikage:
https://github.com/JuliaGraphics/ColorTypes.jl/pull/198#issuecomment-651248852 :sweat_smile:

src/show.jl
```diff
@@ -1,49 +1,32 @@
...
-            $(printargs[:]...)
-        end
-        # Special handling for Normed types: don't print the giant type name
-        function show_normed(io::IO, c::Colorant{FixedPointNumbers.Normed{T,f},$N}) where {T,f}
-            print(io, colorant_string(typeof(c)), '{')
-            FixedPointNumbers.showtype(io, eltype(typeof(c)))
-            print(io, "}(")
-            $(printargs[:]...)
-        end
+        show(io, (comp_n[i])(c))
```
timholy, line 21: Assuming you've checked that this infers properly, I am amazed at how far Julia's compiler has come! (It makes sense that it's possible, but still really cool.)

test/show.jl
```diff
@@ -2,6 +2,20 @@ using ColorTypes
 using ColorTypes.FixedPointNumbers
 using Test
 
+# dummy type for testing 2-component color
+struct AnaglyphColor{T} <: Color{T,2} # not `TransparentGray`
+    left::T; right::T
+end
```
timholy, line 8: Great to see some tests like these. Long on my TODO list.

src/show.jl
kimikage: Sorry to disappoint you, but this does not infer properly. :sweat_smile:
I have determined that the cost of runtime dispatch is negligible, since I did not anticipate the need to print a large number of colors.  However, I will think about the improvement.

timholy: You should be able to fix it with lispy tuple programming, e..g,
```julia
show_foreach(io, c, comp_n...)
```
where
```julia
@inline show_foreach(io, c, f1, f2, f3...) = (show(io, f1(c)); print(io, ','); show_foreach(io, c, f2, f3...))
@inline show_foreach(io, c, f1) = (show(io, f1(c)); print(io, ')'))
```

kimikage: I tried some variations, but the specialization of `show` seems quite complicated. In general, specializations have little effect and most instances are unnecessarily. On the other hand, the specialization for `ARGB32`, for example, is clearly effective.

The current code is not pretty good because we are leaving the specialization to the compiler (not using `@nospecialize`). 
However, there are no fatal problems. I'll do some more study on `show`, but you can also merge this PR.

kimikage:
I am considering changing the display contents themselves in next v0.11 series. (cf. #191, #202, https://github.com/JuliaMath/FixedPointNumbers.jl/issues/188)
Perhaps, the benchmarks and improvements for the current implementation will soon be useless.

So I would like to merge this keeping the implementation simple stupid. If a non-negligible stall is found, we (I) can make a new patch release.

kimikage:
I'll merge this and after a short pause, will release v0.10.6.

PR #187, opened 2020-04-21
Improve `checkval` and `throw_corrorerror`

kimikage:
This changes the first argument of `checkval`/`throw_corrorerror` from a number type to a color type to improve the error message
This also adds `checkval` for 5-component colors.
In some cases (with 2-/4-compornent colors), this slightly improves the performance.

This PR may conflict with the PR #183 (although it doesn't matter) and we should make the decision about https://github.com/JuliaGraphics/ColorTypes.jl/pull/183#issuecomment-616958191.

PR #200, opened 2020-06-27
Add tests for 1-arg rgb constructors

kimikage:
We are going to support `alpha(::Number)` (cf. PR #177).
This adds tests for the 1-arg rgb/transparent-rgb constructors. Some broken tests should be fixed by PR #177.

I'm concerned about supporting 2-arg transparent rgb construcctors. The 2-arg constructors are not the abbreviations for the 3-arg/4-arg constructors.
```julia
julia> ARGB(Gray(1), 0.5)
ARGB{N0f8}(1.0,1.0,1.0,0.502)

julia> ARGB(1, 0.5) # Is this needed?
ERROR: MethodError: no method matching ARGB(::Int64, ::Float64)
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ColorTypes"
 uuid = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
-version = "0.10.4"
+version = "0.10.5"
 
 [deps]
 FixedPointNumbers = "53c48c17-4a7d-5ca2-90c5-79b7896eea93"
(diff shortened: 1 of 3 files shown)


Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "ColorTypes"
 uuid = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
-version = "0.10.4"
+version = "0.10.5"
```
johnnychen94, line 3: I'm unclear about the ColorTypes roadmap. Is this version bump deliberately?

I ask this because this PR doesn't introduce any changes to the `src/` folder.

kimikage: It is intentional, but not definitive. It was just moved from PR #187.

Issue #202, opened 2020-07-03
[RFC] Simplified display of colors and arrays of colors


kimikage:
This is similar to issue #191, but for non-gray types.

If we think of colors as containers for numbers, the following is also reasonable and consistent with `Array`:
(current)`"RGB{Float32}(1.0f0,0.5f0,0.0f0)"`-->(proposed)`"RGB{Float32}(1.0,0.5,0.0)"`

In fact, `Normed` colors have that kind of special handling.
```julia
julia> RGB{N0f8}(1,0.5,0)
RGB{N0f8}(1.0,0.502,0.0)
```
Of course, the following are completely different in implementation, but I think that the second style has no problem in "display".
```julia
RGB{Float32}[RGB{Float32}(1.0f0,0.5f0,0.0f0)] # fully specified (current)

RGB{Float32}[RGB(1.0,0.5,0.0)]                # via RGB{Float64} (proposed)

RGB{Float32}[RGB{Float32}(1.0,0.5,0.0)]       # without :typeinfo handling

RGB{Float32}[RGB(1.0f0,0.5f0,0.0f0)]          # bottom-up and top-down are confusing
```

johnnychen94:
If the colorant constructor accepts tuple as input, then we may get an even concise display:

```julia
RGB{Float32}[(1.0,0.5,0.0)] # if RGB((1.0, 0.5, 0.0)) works
```

kimikage:
I was thinking the same thing.
Perhaps in order for the color constructors to accept `Tuple` inputs, the color to `Tuple` conversions should be implement first.(cf. https://github.com/JuliaGraphics/ColorVectorSpace.jl/pull/131#discussion_r443177446) That is, we have to make a strict distinction between a sequence of constructor arguments (i.e. something like `Vararg`) and a tuple representation of a color.

In any case, we should report separate issues.

johnnychen94:
> That is, we have to make a strict distinction between a sequence of constructor arguments (i.e. something like Vararg) and a tuple representation of a color.

I'm not sure if I get your concerns clearly; are there any conceptual differences between `RGB(1.0, 0.5, 0.0)` and `RGB((1.0, 0.5, 0.0))`, or will it introduce any ambiguity here?

kimikage:
I think the conceptual difference is essentially the first case of https://docs.julialang.org/en/v1/manual/conversion-and-promotion/#Conversion-vs.-Construction-1.

[...]

PR #203, opened 2020-07-04
Version 0.10.6

kimikage:
This also modifies the test for Julia nightly build.

### DIFF
--- a/test/show.jl
+++ b/test/show.jl
@@ -70,14 +70,29 @@ end
 @testset "collection of colors" begin
     iob = IOBuffer()
     summary(iob, Gray{N0f8}[0.2, 0.4, 0.6])
-    @test String(take!(iob)) == "3-element Array{Gray{N0f8},1} with eltype Gray{Normed{UInt8,8}}"
+    vec_summary = String(take!(iob))
+
     mat = RGB{Float64}[RGB(1,0,0) RGB(0,1,0); RGB(0,0,1) RGB(1,1,1)]
     summary(iob, mat)
-    @test String(take!(iob)) == "2×2 Array{RGB{Float64},2} with eltype RGB{Float64}"
+    mat_summary = String(take!(iob))
+
     summary(iob, view(mat,:,:))
-    @test String(take!(iob)) == "2×2 view(::Array{RGB{Float64},2}, :, :) with eltype RGB{Float64}"
+    view_summary = String(take!(iob))
+
     summary(iob, TransparentColor[ARGB32(), HSVA(30,1,1,0.5)])
-    @test String(take!(iob)) == "2-element Array{TransparentColor,1} with eltype TransparentColor"
+    avec_summary = String(take!(iob))
+
+    if VERSION >= v"1.6.0-DEV.356"
+        @test_broken vec_summary == "3-element Vector{Gray{N0f8}}"
+        @test_broken mat_summary == "2×2 Matrix{RGB{Float64}}"
+        @test_broken view_summary == "2×2 view(::Matrix{RGB{Float64}}, :, :) with eltype RGB{Float64}"
+        @test_broken avec_summary == "2-element Vector{TransparentColor}"
+    else
+        @test vec_summary == "3-element Array{Gray{N0f8},1} with eltype Gray{Normed{UInt8,8}}"
+        @test mat_summary == "2×2 Array{RGB{Float64},2} with eltype RGB{Float64}"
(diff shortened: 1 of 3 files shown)


test/show.jl
```diff
@@ -70,14 +70,29 @@ end
...
-    @test String(take!(iob)) == "2-element Array{TransparentColor,1} with eltype TransparentColor"
+    avec_summary = String(take!(iob))
+
+    if VERSION >= v"1.6.0-DEV.356"
+        @test_broken vec_summary == "3-element Vector{Gray{N0f8}}"
+        @test_broken mat_summary == "2×2 Matrix{RGB{Float64}}"
+        @test_broken view_summary == "2×2 view(::Matrix{RGB{Float64}}, :, :) with eltype RGB{Float64}"
+        @test_broken avec_summary == "2-element Vector{TransparentColor}"
+    else
+        @test vec_summary == "3-element Array{Gray{N0f8},1} with eltype Gray{Normed{UInt8,8}}"
```
johnnychen94, line 91: We can use `typeof(a)` here instead to avoid this version check.

See also: https://github.com/JuliaArrays/OffsetArrays.jl/pull/125

test/show.jl
```diff
@@ -104,7 +119,12 @@ end
     @test ColorTypes.colorant_string_with_eltype(RGB{Union{}}) == "RGB{Union{}}"
     @test ColorTypes.colorant_string_with_eltype(RGB{<:Fractional}) == "RGB"
     @test ColorTypes.colorant_string_with_eltype(HSV{<:AbstractFloat}) == "HSV"
-    @test ColorTypes.colorant_string_with_eltype(RGB{Fractional}) == "RGB{Union{AbstractFloat, FixedPoint}}"
+    rgb_fractional = ColorTypes.colorant_string_with_eltype(RGB{Fractional})
+    if VERSION >= v"1.6.0-DEV.356"
+        @test eval(Meta.parse(rgb_fractional)) == RGB{Union{AbstractFloat, FixedPoint}}
+    else
+        @test rgb_fractional == "RGB{Union{AbstractFloat, FixedPoint}}"
```
johnnychen94, line 126: and `@test rgb_fractional == "RGB{$(Fractional)}"`, too?

test/show.jl
kimikage: That's right, but that looks like a replica of the implementation. :confused:

johnnychen94: I don't see if there is any issue as long as it is consistent with the Base implementation.

johnnychen94: Seems not a problem to me. It doesn't reuse any show methods defined in this package.

kimikage: These are the tests for `showarg` rather than `summary`.
My concern is the quality of the test, not the implementation. There are two different points of view:
- we have succeeded in detecting the specification change
- we have failed to follow the specification change

johnnychen94: There's no doubt that it gives better test quality to track every specific change. However, for IO changes, IMO it's really exhausting to track down every detail. So as a tradeoff between maintainability and test quality, I feel it okay as long as it follows a same pattern to Base.

Of course, this is just my preference.

test/show.jl
kimikage: Unlike the cases of `summary`, this test aims to clarify the current behavior. The important thing is that it is "not" `"RGB{Fractional}"`.
In this `Fractional` case, the results on the nightly build is still fragile, so I decided to check for the minimal round-tripness.

test/show.jl
kimikage: Note that there is no "with eltype" for `Array` in `Base`. "with eltype N0f8" is not wrong, but it is not the expected result.

johnnychen94: I think you might misunderstand my solution:

```julia
iob = IOBuffer()
avec = Gray{N0f8}[0.2, 0.4, 0.6]
vec_summary = summary(iob, avec)

...

@test vec_summary == "3-element $(typeof(avec)) with eltype Gray{Normed{UInt8,8}}"
```

kimikage: I'm sorry, I was assuming you were suggesting `eltype with $(eltype(avec))`.
If so, what are you going to test? :joy:

kimikage:
In my fork, the Julia nightly test on Linux was passed.
Regarding the display of colors, I will soon change the behavior (cf. #191, #202). So the tests will be changed, too.

PR #204, opened 2020-07-04
CI updates

johnnychen94:
I believe it's an oversight to not test Windows?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: julia
 os:
   - linux
   - osx
+  - windows
 julia:
   - 1.0
   - 1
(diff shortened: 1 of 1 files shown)


kimikage:
I do not object to adding Windows. (Since I mainly develop the ColorTypes on Windows, it doesn't bother me even if we don't add it, though. :smile:)

We haven't checked on 32-bit or ARM systems. In terms of "finding bugs", I think it is more beneficial to add those environments. On the other hand, in terms of covering more users' environments, adding Windows is good.

I guess @timholy is most concerned about CI time.

johnnychen94:
> I guess @timholy is most concerned about CI time.

In that case, we can switch to Github action, which enables more parallel jobs.

PR #209, opened 2020-07-23
Reimplement `parameter_upper_bound`

### LINKED ISSUE #208: `Base.parameter_upper_bound` was removed in Julia v1.6-DEV

kimikage:
Fixes #208

This reimplements the function, although the implementation still depends on non-public APIs (i.e. `Base.unwrap_unionall` and `Base.rewrap_unionall`).
This also generalizes `eltypes_supported` for non-parametric color types.

We haven't merged the breaking changes yet, so I plan to release v0.10.7.

### DIFF
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -129,13 +129,16 @@ eltype(::Type{Colorant{T,N}}) where {T,N}   = T
 
 eltype(c::Colorant) = eltype(typeof(c))
 
+@pure function _parameter_upper_bound(t::UnionAll, idx)
+    Base.rewrap_unionall((Base.unwrap_unionall(t)::DataType).parameters[idx], t)
+end
+
 # eltypes_supported(Colorant{T<:X}) -> X
-@pure eltypes_supported(::Type{C}) where {C<:Colorant} =
-    Base.parameter_upper_bound(base_colorant_type(C), 1)
-eltypes_supported(::Type{RGB24})   = N0f8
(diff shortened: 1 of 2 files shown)


src/traits.jl
```diff
@@ -129,13 +129,16 @@ eltype(::Type{Colorant{T,N}}) where {T,N}   = T
 
 eltype(c::Colorant) = eltype(typeof(c))
 
+@pure function _parameter_upper_bound(t::UnionAll, idx)
```
vtjnash, line 132: It can be risky to claim a function is supposedly `@pure`. This was deprecated in Base due to this issue, and because the type system can now handle such queries correctly through dispatch(*), while `similar` already implements what's often meant, but better.

(*) The mistake often made in using this function is the assumption that subtype parameters are of the same number and order as their supertype. This happens to be true inside this package for some of its types, but is easily seen to be false for other examples like `BitArray <: AbstractArray{Bool}`. Thus, the correct solution is usually to ask dispatch for the matching parameters of the _supertype_, like so:
```
eltype(::Type{<:AbstractArray{T}}) where {T} = T
```

kimikage: > The mistake often made in using this function is the assumption that subtype parameters are of the same number and order as their supertype. 

I understand. However, that assumption seems necessary. I think it is the correct behavior to get the wrong result with the wrong usage.

PR #210, opened 2020-07-24
Suppress display of module names for invisible types

kimikage:
In PR #174 (v0.10.0), the module names were displayed for invisible non-concrete types. In PR #199, non-parametric types also joined the problem. This suppresses the display of module names.

This also completely fixes the problem with display of `AGray32`. (It was accidentally half-fixed in PR #199.)

```julia
julia> import ColorTypes # not `using`

julia> ColorTypes.RGB[] # v0.10.0
0-element Array{ColorTypes.RGB,1} with eltype ColorTypes.RGB

julia> ColorTypes.RGB24[] # v0.10.6
0-element Array{ColorTypes.RGB24,1} with eltype ColorTypes.RGB24

julia> ColorTypes.RGB24[] # this PR
0-element Array{RGB24,1} with eltype ColorTypes.RGB24
```
```julia
julia> AGray32(1) # v0.10.0 or earlier
AGray32{N0f8}(1.0,1.0)

julia> AGray32(1) # v0.10.6 Damn!
AGray32(1.0,1.0)

julia> AGray32(1) # this PR
AGray32(1.0N0f8,1.0N0f8)
```

PR #211, opened 2020-07-31
Make use of `_parameter_upper_bound` safer

### LINKED ISSUE #208: `Base.parameter_upper_bound` was removed in Julia v1.6-DEV

kimikage:
This makes sure that `_parameter upper_bound` is used with `TransparentColor`, not its subtypes.
This also changes the implementation of `eltypes_supported`.

Closes #208

### DIFF
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -137,7 +137,11 @@ end
 function eltypes_supported(::Type{C}) where {C<:Colorant}
     Cb = base_colorant_type(C)
     isconcretetype(C) && C === Cb && return eltype(C)
-    _parameter_upper_bound(Cb, 1)
+    _eltypes_supported(Cb)
+end
+@pure function _eltypes_supported(::Type{C}) where {C<:Colorant}
+    nameof(C) === :Colorant || return _eltypes_supported(supertype(C))
+    _parameter_upper_bound(C, 1)
 end
 
 eltypes_supported(c::Colorant) = eltypes_supported(typeof(c))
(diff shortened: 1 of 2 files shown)


src/traits.jl
```diff
@@ -129,15 +129,16 @@ eltype(::Type{Colorant{T,N}}) where {T,N}   = T
...
 # eltypes_supported(Colorant{T<:X}) -> X
 function eltypes_supported(::Type{C}) where {C<:Colorant}
     Cb = base_colorant_type(C)
     isconcretetype(C) && C === Cb && return eltype(C)
-    _parameter_upper_bound(Cb, 1)
+    _eltypes_supported(Cb)
+end
+@pure function _eltypes_supported(::Type{C}) where {C<:Colorant}
+    nameof(C) === :Colorant || return _eltypes_supported(supertype(C))
+    C.var.name === :T || return Any
```
timholy, line 140: Depending on `C.var.name` seems a little risky, though surprisingly it works for now:
```julia
julia> abstract type Container{T} end

julia> foo(::Type{C}) where C<:Container = C.var.name
foo (generic function with 1 method)

julia> foo(Container)
:T

julia> foo(Container{U} where U)
:T

julia> C = Container{U} where U
Container{U} where U

julia> C.var.name
:U
```

test/traits.jl
```diff
@@ -224,11 +233,25 @@ end
...
+    @test @inferred(!ColorTypes.issupported(AGray, Bool))
+
+    @test @inferred(ColorTypes.eltypes_supported(C2{Float32})) === Real
+    @test_broken @inferred(ColorTypes.eltypes_supported(C2A)) === Real
+
+    @test @inferred(ColorTypes.eltypes_supported(TransparentColor)) === Any
+
+    ST = @inferred(ColorTypes.eltypes_supported(StrangeGray{:X}))
+    @test_broken ST === (Normed{T} where T <: Integer)
+    @test_broken eltype(StrangeGray{UInt16}(1N14f2)) isa ST
```
timholy, line 251:
```suggestion
    @test_broken eltype(StrangeGray{UInt16}(1N14f2)) <: ST
```
but I think this will pass?

kimikage: I intended to change the `eltype` to `gray`. It seems I forgot that while doing other work. :sweat_smile:
In any case, this is a `broken` for now.

src/traits.jl
kimikage: Yes, this is a makeshift solution. I'm looking for a better way this past week.

kimikage: I changed the code to use `_parameter_upper_bound` again. However, the new implementation makes sure the input is a `Colorant`, so it is safer and more reasonable than the current master.

```julia
julia> ColorTypes.eltypes_supported(TransparentColor)
Color # current master
Any   # this PR

julia> ColorTypes.eltypes_supported(TransparentColor{RGB{T},T,4} where T) # analysis of constraints is still loose.
Any
```

kimikage:
Thank you for the review. The last `push -f` is due to trivial code formatting.

I will merge this and #210, and then release v0.10.8. I believe it will be the last release of the v0.10 series, for sure. :sweat_smile:

PR #212, opened 2020-08-13
Simplify the implementation of eltype

timholy:
I'm guessing the old implementation predated Julia 0.6 and the new type system.

### DIFF
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -123,9 +123,7 @@ to_top(::Type{Colorant{T,N}}) where {T,N} = Colorant{T,N}
 to_top(c::Colorant) = to_top(typeof(c))
 
 # eltype(RGB{Float32}) -> Float32
-eltype(::Type{Colorant{T}}) where {T}       = T
-eltype(::Type{Colorant{T,N}}) where {T,N}   = T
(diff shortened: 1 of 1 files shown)


timholy:
Yep, still returns `Any`. `eltype(RGB)` dispatches to [this method](https://github.com/JuliaLang/julia/blob/fedde78fe610b4f1fad25b723091489b4604c3ab/base/array.jl#L123). If that seems OK to you, then I don't think this changes behavior.

Alternatively, the implementation
```julia
eltype(::Type{C}) where {C<:Colorant{T}} where {T} = T
eltype(::Type{C}) where {C<:Colorant} = Base.parameter_upper_bound(C, 1)
```
returns the "valid" type as implied in #151. Thoughts?

kimikage:
> oh, I see I even mentioned that in #151 (comment), but I don't think we really decided.

Yes, and we now have the problem with `parameter_upper_bound`(cf. https://github.com/JuliaGraphics/ColorTypes.jl/issues/208). If we use `_parameter_upper_bound`, the recursion still seems to be a necessary technique.

Until we make a decision about PR #184, we should not change the current behavior.

timholy:
Interestingly, this implementation is much friendlier to precompilation of ImageCore methods.

PR #214, opened 2020-08-27
Add a few more precompiles

timholy:
These precompiles help ImageCore precompile other MethodInstances.

### DIFF
--- a/src/precompile.jl
+++ b/src/precompile.jl
@@ -59,6 +59,8 @@ function _precompile_()
     end
     # traits
     for T in eltypes, C in pctypes
+        @assert precompile(Tuple{typeof(eltype),Type{C}})
+        @assert precompile(Tuple{typeof(eltype),Type{C{T}}})
         @assert precompile(Tuple{typeof(alpha),C{T}})
         C <: AbstractGray && @assert precompile(Tuple{typeof(gray),C{T}})
         C <: AbstractRGB || continue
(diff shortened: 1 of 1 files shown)


src/precompile.jl
```diff
@@ -59,6 +59,8 @@ function _precompile_()
     end
     # traits
     for T in eltypes, C in pctypes
+        @assert precompile(Tuple{typeof(eltype),Type{C}})
+        @assert precompile(Tuple{typeof(eltype),Type{C{T}}})
         @assert precompile(Tuple{typeof(alpha),C{T}})
         C <: AbstractGray && @assert precompile(Tuple{typeof(gray),C{T}})
         C <: AbstractRGB || continue
```
kimikage, line 66: This is not directly related to this PR, but are the `red()`, `green()` and `blue()` for `TransparentRGB` not precompiled?

kimikage (approved):
I haven't quantified the benefits, but at least the precompilation of ColorTypes isn't too slow.

src/precompile.jl
timholy: No, I could do that but I think transparent colors are less commonly-used. And of course having them for `RGB` saves some time when it does go to compile them.

timholy:
Benefits are not huge but it helps other packages. There's a chain from ColorTypes->ImageCore->PNGFiles, and `load(filename)` takes 0.2s to infer which is a substantial time.

Issue #150, opened 2020-01-15
[RFC] Dropping support for non-RGB/Gray types in `rand`


kimikage:
Probably the next release of Colors.jl will depend "tightly" on the next release of ColorTypes.jl. So, what about moving `rand` for color types to Colors.jl on this occasion?

cf. https://github.com/JuliaGraphics/ColorTypes.jl/pull/140

**Edit:**
ColorTypes v0.10 and Colors v0.12 were released without changing `rand`. I will leave this issue open for the time being.

timholy:
Does anyone besides me feel a bit uncomfortable about supporting `rand` more broadly across color types? What bothers me is the sampling. For example, suppose you want to sample points inside the circle. You might think that the following two approaches would yield the same result:
```julia
julia> function rejection(n)
           pts = Tuple{Float64,Float64}[]
           while length(pts) < n
               x, y = 2*rand()-1, 2*rand()-1
               if x^2 + y^2 <= 1
                   push!(pts, (x, y))
               end
           end
           return pts
       end
rejection (generic function with 1 method)

julia> function polar(n)
           function randpolar()
               r, θ = rand(), 2π*rand()
               return (r*cos(θ), r*sin(θ))
           end
           return [randpolar() for i = 1:n]
       end
```
Yet if you generate 10^4 samples from each and plot the histogram of `x` values, you get this:
![hist](https://user-images.githubusercontent.com/1525481/72609404-6965bf80-38ea-11ea-96af-ff30f74340e9.png)

As a consequence, doing `rand(RGB{Float32}, n)` and then converting to HSV will lead to a different result than `rand(HSV{Float32}, n)`.

Perhaps we could just document this. But I thought I'd at least mention that it bothers me a bit.

kimikage:
> As a consequence, doing `rand(RGB{Float32}, n)` and then converting to HSV will lead to a different result than `rand(HSV{Float32}, n)`.

I personally think that is obvious. In the first place, I think the problem is to use a uniform distribution without recognizing those differences. :confused:
If my feeling is far from many people, I have to be more careful to support the gamut. In general, the gamut is not a 2-D figure or 3-D cube, but a complex hull or a set of multiple hulls. I proposed using a bounding box to avoid that complexity (cf. #140). However, `rand` is dangerous because it is too easy to use.

kimikage:
[...]

PR #221, opened 2021-01-23
Change line ending code of "test/operations.jl" and refactor `rand` tests

kimikage:
This adds the broken test to identify the problems with `rand`.

cf. https://github.com/JuliaGraphics/ColorTypes.jl/issues/150#issuecomment-691455695, https://github.com/JuliaGraphics/ColorTypes.jl/issues/150#issuecomment-691459180
```julia
julia> rand(AHSV) # transparent color except for RGB and Gray
ERROR: MethodError: no method matching gamutmax(::Type{AHSV{Float64}})

julia> rand(RGB24)
ERROR: TypeError: in Type{...} expression, expected UnionAll, got Type{RGB24}

julia> rand(ARGB32)
ERROR: TypeError: in Type{...} expression, expected UnionAll, got Type{ARGB32}

julia> using Random, FixedPointNumbers

julia> rand(MersenneTwister(), RGB{Float32}) # RNG option
ERROR: ArgumentError: Sampler for this object is not defined

julia> rand(RGB{Q1f6}) # invalid range with (signed) `Fixed`
RGB{Q1f6}(0.23Q1f6,-0.41Q1f6,-0.12Q1f6)
```


kimikage:
I am going to merge this in the next few days, since this does not change the "src", and it is better to fix the line ending code as early as possible.

PR #222, opened 2021-01-23
[RFC] Use sampler-based Random API

kimikage:
This is a proposal to drastically change the implementation of `rand`. (cf. https://github.com/JuliaGraphics/ColorTypes.jl/issues/150#issuecomment-594986636)

### DIFF
--- a/src/operations.jl
+++ b/src/operations.jl
@@ -50,79 +50,103 @@ gamutmax(::Type{T}) where {T<:HSL} = (360,1,1)
 gamutmin(::Type{T}) where {T<:HSL} = (0,0,0)
 gamutmax(::Type{T}) where {T<:Lab} = (100,128,128)
 gamutmin(::Type{T}) where {T<:Lab} = (0,-127,-127)
-gamutmax(::Type{T}) where {T<:LCHab} = (100,1,360)
+gamutmax(::Type{T}) where {T<:LCHab} = (100,1,360) # FIXME
 gamutmin(::Type{T}) where {T<:LCHab} = (0,0,0)
-gamutmax(::Type{T}) where {T<:YIQ} = (1,0.5226,0.5226)
-gamutmin(::Type{T}) where {T<:YIQ} = (0,-0.5957,-0.5957)
+gamutmax(::Type{T}) where {T<:YIQ} = (1,0.5226,0.5226) # FIXME
+gamutmin(::Type{T}) where {T<:YIQ} = (0,-0.5957,-0.5957) # FIXME
 
 gamutmax(::Type{T}) where {T<:AbstractGray} = (1,)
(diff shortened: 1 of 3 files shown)


kimikage:
At this time, I have not applied the `rand!` optimization using `reinterpret` or `unsafe_wrap`, but it is not fatally slow.
```julia
julia> @btime rand(RGB{Float32}, 1000, 1000, 10);
  97.308 ms (2 allocations: 114.44 MiB) # v1.6.0-beta1 Windows
  95.061 ms (2 allocations: 114.44 MiB) # v1.6.0-beta1 Debian on WSL2
  134.923 ms (2 allocations: 114.44 MiB) # v1.5.3 Windows

julia> @btime rand(HSV{Float32}, 1000, 1000, 10);
  124.078 ms (2 allocations: 114.44 MiB) # v1.6.0-beta1 Windows
  122.759 ms (2 allocations: 114.44 MiB) # v1.6.0-beta1 Debian on WSL2
  152.048 ms (2 allocations: 114.44 MiB) # v1.5.3 Windows
```

kimikage:
I'm not sure the reason, but precompilation seems to be of little help.

src/operations.jl
```diff
@@ -13,80 +13,105 @@ gamutmax(::Type{T}) where {T<:HSL} = (360,1,1)
...
+    rand!(r, A, SamplerType{base_colorant_type(C){Float64}}())
+end
+function rand!(r::AbstractRNG, A::Array{C}, ::SamplerType{C}) where {T, C<:Colorant{T}}
+    A .= rand.((r,), C)
+end
+function _rand01!(r::AbstractRNG, A::Array{C},
+               ::SamplerType{C}) where {T<:Rand01Type, C<:Colorant{T}}
+    N = sizeof(C) ÷ sizeof(T)
+    T0 = T <: FixedPoint ? FixedPointNumbers.rawtype(T) : T
+    At = unsafe_wrap(Array, reinterpret(Ptr{T0}, pointer(A)), (N, size(A)...))
```
timholy, line 79: Are you using `unsafe_wrap` to bypass the performance problem of `ReinterpretArray` prior to Julia 1.6? I'm fine with that, I just remember someone complaining about that solution previously.

timholy (approved):
I haven't reviewed this in gory detail, but it looks like you're adding more flexibility by supporting different samplers, which seems like a win to me and a very nice change.

src/operations.jl
kimikage: No, it has nothing to do with `ReinterpretArray` here. (Of course, therein lies the root problem.)
It is an issue with using the private API.
https://github.com/JuliaLang/julia/issues/36718

kimikage:
In any case, this is too big a change to include in a patch release.

PR #223, opened 2021-02-06
Added Color to Array convert functions and tests

cfitz25:
Created convert functions to turn ColorTypes into Arrays, as most languages i have used typically treat colours as arrays i thought it would be useful to allow the same with Julia.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 *.jl.cov
 *.jl.mem
 Manifest.toml
+.vscode/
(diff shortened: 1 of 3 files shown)


kimikage:
I agree that this kind of conversion can be useful in specific use cases. However, in the Julia or JuliaImages ecosystem, it is not necessarily a best practice.

If you "really" need the array, it is better to construct it explicitly. As long as the `convert` is called directly by the user, I don't think there will be much confusion, but I'm not sure if it will be consistent when called from other methods implicitly.

cfitz25:
Any ideas what functions may break from this? I did try with converting 2d arrays of RGB to a 3d array (or more a 2d array of 1d arrays) and that appeared to work correctly atleast.

timholy:
There's an ambiguity about whether you want to `reinterpret` or use the constructor order. Consider `ARGB`: the `A` comes first in the list and memory order, but when constructing colors we always have the user specify the alpha channel last.

ImageCore has utilities for this: it exports `reinterpretc` and `channelview`, so you have explicit control over these things.

cfitz25:
@timholy thank you for pointing me in the right direction

timholy:
OK to close this?

PR #206, opened 2020-07-06
Simplify display of colors and arrays of colors

kimikage:
Fixes #202, Fixes #191 ~(and https://github.com/JuliaGraphics/ColorTypes.jl/issues/179#issuecomment-617131514)~

This simplifies the display of array elements based on the `:typeinfo` property. This also simplifies the display of color components by setting the `:typeinfo` property.
This prints the elements of gray arrays as real numbers.

PR #227, opened 2021-02-25
Version 0.10.10

kimikage:
This is the last checkpoint for the release of ColorTypes v0.11. (I think I've declared something similar in the past. :sweat_smile:)

The following is the correspondence between the versions I had assumed.
|ColorTypes      | v0.10.10 | v0.11.0 (#213, #186?) | v0.12.0 (#177, #206, #222, #197?)|
|:---------------|:---------|:----------------------|:--------|
|Colors          |(v0.12.6) | v0.12.7               | v0.13.0 |
|ColorVectorSpace|(v0.9.1)  | v0.9.2                | v0.9.3  |

The important thing is that #222 will break a lot of doctests, so have it guarded by Colors.jl's cap. So I am concerned about the end of ColorTypes v0.11, i.e. the start of ColorTypes v0.12. (https://github.com/JuliaGraphics/ColorTypes.jl/pull/213#issuecomment-673599021)
There is room for debate as to what to include in the ColorTypes v0.11 series, but I think it is straightforward to limit it to only ColorVectorSpace related items. They do not affect Colors. (https://github.com/JuliaGraphics/ColorTypes.jl/pull/213#issuecomment-673576572)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ColorTypes"
 uuid = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
-version = "0.10.9"
+version = "0.10.10"
 
 [deps]
 FixedPointNumbers = "53c48c17-4a7d-5ca2-90c5-79b7896eea93"


kimikage:
@timholy, do you plan to refactor #186 and include it in v0.11.0?

timholy:
Yes, once you're content with 0.10 I'd like to do some of those kinds of cleanups for more sensible integration with ColorVectorSpace.

timholy (approved):
0.10 has been great, let's do 0.11 soon!

kimikage:
Let's release v0.10.10, and if it doesn't seem to affect downstream packages, then merge #213 this weekend.

kimikage:
Oh, I just found out about the FileIO precompilation problem. :sweat_smile:
Whatever it is, let's see how it goes for a moment.

timholy:
It's fixed if you upgrade to FileIO 1.6. You also want Revise 3.1.14.

timholy:
Only affects Julia 1.7 within the last 36 hours.

kimikage:
According to PkgEval and some local tests, there seems to be no problem with v0.10.10 itself.
However, I have noticed a regression in v0.10. https://github.com/JuliaGraphics/ColorTypes.jl/issues/208#issuecomment-792131700

PR #228, opened 2021-03-03
Generalize `==`

kimikage:
cf. https://github.com/JuliaGraphics/ColorTypes.jl/pull/186#issuecomment-672044220

This would qualify as part of v0.10.x, but in downstream packages, the benefits of the generalization would have to wait until v0.11.0. I'm submitting this PR now because I'm assuming we'll be moving the comparison methods to "operations.jl" or "comparisons.jl" (or something). In other words, this is a decluttering before the move.


timholy (approved):
:+1:

kimikage:
This does not conflict with #213 and is not directly related to ColorVectorSpace. So, I would like to include this in v0.10.10.

PR #232, opened 2021-03-04
Warn, not error, on precompile failure

timholy:


kimikage:
Thank you for addressing this potential risk.:+1:

However, where is the original code of `@warnpcfail`? In other words, where should the discussion about it take place? `SnoopCompile`?

As a practical matter, doesn't this macro generate a warning on every call, causing a slowdown? :confused:

timholy:
It only runs when the package is being built (aka, precompiled). The reason the coverage looks so low is because the macro code doesn't even show up as having been executed when the tests are run.

I agree it's not ideal to have to define the macro in each package, but we don't seem to have a good alternative. So yes, file issues in SnoopCompile.

kimikage:
> It only runs when the package is being built (aka, precompiled).

That's right. However, developers and CI go through precompilation many times. I don't think it's a good idea to spend time on a feature which will basically never be used.

timholy:
So the three of us who contribute to this package should optimize our own welfare at the expense of the ~150K Julia users?

Anything that's obviously silly to precompile should be expunged. But I don't see obvious candidates. Besides, that's why I wrote Revise.

kimikage:
I may have misled you. I am talking just about this PR.

kimikage:
I mean, I think we should have moved `@warn` to a `@noinline` function.:grapes:

timholy:
Happy to accept a PR!

kimikage:
In terms of performance, simply kicking out @warn seems to be about the same as @assert version. However, I think it is better to review more fundamentally.
I'd like to open an issue on the matter first.

BTW, now that the problem with FileIO has been solved, the problem with PNGFiles seems to have become apparent.

timholy:
I'm happy to wait for you to investigate.

> the problem with PNGFiles seems to have become apparent

Reference? (I haven't seen this though perhaps it's sitting in my inbox and I will later this morning)

kimikage:
Sorry for the misinformation. It was ReferenceTests, not PNGFiles (and the problem was already fixed by you :+1:).
There is a potential problem with `@assert` in PNGFiles as well, though.

kimikage:
xref: https://github.com/timholy/SnoopCompile.jl/issues/243

kimikage:
I'm not strongly opposed to merging this PR. However, to be honest, having experienced precompilation many times in PR #233, I do not welcome the precompilation to be unnecessarily slow.

kimikage:
So, is it enough to simply remove `@assert`s?

kimikage:
In this case, I chose PR #234 for performance reasons, but that is not to dismiss more exhaustive checks and polite warning messages.

src/precompile.jl
```diff
@@ -1,3 +1,14 @@
+macro warnpcfail(ex::Expr)
+    modl = __module__
+    file = __source__.file === nothing ? "?" : String(__source__.file)
+    line = __source__.line
+    quote
+        $(esc(ex)) || @warn """precompile directive
```
johnnychen94, line 6: Sorry for the late feedback. Maybe a `@debug` instead of `@warn` can be more helpful when someone wants to investigate whether the precompilations are still valid?

kimikage: Did you mean https://github.com/JuliaLang/julia/pull/41447?

johnnychen94: Exactly! And doing it upstream is better than doing it here.

PR #233, opened 2021-03-07
Fix regression of constant propagation in `color_type`

### LINKED ISSUE #208: `Base.parameter_upper_bound` was removed in Julia v1.6-DEV

kimikage:
The recursion mechanism using `supertype` could be broken when a subtype of `TransparentColor` was added. Therefore, the checking with `nameof` was used for safety in PR #211.
However, it interfered with constant folding. This avoids the use of `nameof` for the checking.

Closes #208

### DIFF
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -135,12 +135,11 @@ end
 function eltypes_supported(::Type{C}) where {C<:Colorant}
     Cb = base_colorant_type(C)
     isconcretetype(C) && C === Cb && return eltype(C)
-    _eltypes_supported(Cb)
+    _eltypes_supported(Cb, supertype(Cb))
 end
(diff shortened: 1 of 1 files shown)


kimikage:
```julia
julia> @btime color_type(ARGB)(g) setup=(g=rand());
  28.083 ns (0 allocations: 0 bytes) # v0.10.8-v0.10.10
  1.099 ns (0 allocations: 0 bytes)  # this PR
```

kimikage:
If we are going to merge PR #232 as it stands, or if we can make an improved version soon, I will release v0.10.11 with this and PR #232.

kimikage:
Julia's compiler is too hard to understand. :confused:

kimikage:
I think this is potentially fragile. However, since its negative effect is a failure of inference and constant propagation, I think solving the problem we know now is more important than the potential problems.

PR #234, opened 2021-03-12
Remove assertions in "src/precompile.jl"

kimikage:
The bit rot in this package itself should be detected by testing with Julia v1.7.0-DEV.694 or later.

This is an alternative to PR #232.


timholy (approved):

kimikage:
Now I will merge them and register v0.10.11. These two commits should not affect the superficial behavior, so if PkgEval does not show any problems, I think we can move on.

kimikage:
Hmmm, there is still a problem with #233 :confused:

PR #236, opened 2021-03-13
Fix inference regression on `eltypes_supported`

kimikage:
I am not fully aware of this inference issue, but v0.10.11 affects the testing of ImageCore.jl, so I am releasing this as a quick fix.


kimikage:
Since `eltypes_supported` is difficult for the compiler to infer, the `@generated` function seems to be more suitable (using loops instead of recursion).

kimikage:
Incidentally, from this time on, it is being tested with Julia v1.5.4; the testing of some of the downstream packages with Julia v1.5.3 was done locally.

PR #238, opened 2021-03-15
Add `zero` & `oneunit` for more types

### LINKED ISSUE #219: ERROR: MethodError: no method matching zero(::Gray{Float64})
Apparently, `Gray.(A)` is only defined for some matrix types, but not for, e.g., `LinearAlgebra.UpperTriangular` matrices:

```julia
julia> using LinearAlgebra, Colors

julia> A = rand(3, 3)
3×3 Array{Float64,2}:
 0.717871  0.265517   0.215277
 0.901667  0.0204098  0.358058
 0.583615  0.47894    0.42068

julia> Gray.(A)  # works fine
3×3 Array{Gray{Float64},2} with eltype Gray{Float64}:
 Gray{Float64}(0.717871)  Gray{Float64}(0.265517)   Gray{Float64}(0.215277)
 Gray{Float64}(0.901667)  Gray{Float64}(0.0204098)  Gray{Float64}(0.358058)
 Gray{Float64}(0.583615)  Gray{Float64}(0.47894)    Gray{Float64}(0.42068)

julia> Gray.(UpperTriangular(A))
3×3 UpperTriangular{Gray{Float64},Array{Gray{Float64},2}}:
Error showing value of type UpperTriangular{Gray{Float64},Array{Gray{Float64},2}}:
ERROR: MethodError: no method matching zero(::Gray{Float64})
```

```julia
(@v1.4) pkg> status Colors
Status `~/.julia/environments/v1.4/Project.toml`
  [5ae59095] Colors v0.12.6
```

kimikage:
This is a part of PR #186.

This moves the `zero` and `oneunit` for  some of grays and rgbs that were defined in ColorVectorSpace to ColorTypes. (That is, this is a breaking change.)
For `zero`, a color with all components zero is considered reasonable as an additive identity in most color spaces. On the other hand, it is difficult to define `oneunit` in general. Therefore, `oneunit` is defined only for specific color spaces.

This does not change the definition of `one`.

Fixes #219

### DIFF
--- a/src/error_hints.jl
+++ b/src/error_hints.jl
@@ -2,10 +2,10 @@
 function register_hints()
     if isdefined(Base, :Experimental) && isdefined(Base.Experimental, :register_error_hint)
         Base.Experimental.register_error_hint(MethodError) do io, exc, argtypes, kwargs
-            if exc.f in (zero, one) && argtypes[1] <: Union{Type{<:AbstractRGB}, AbstractRGB}
+            if exc.f in (one,) && argtypes[1] <: Union{Type{<:AbstractRGB}, AbstractRGB}
                 print(io, "\nYou may need to `using ColorVectorSpace`.")
             end
-            if exc.f in (zeros, ones) && argtypes[1] <: Type{<:AbstractRGB}
+            if exc.f in (ones,) && argtypes[1] <: Type{<:AbstractRGB}
                 print(io, "\nYou may need to `using ColorVectorSpace`.")
             end
         end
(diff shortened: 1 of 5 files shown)


kimikage:
In addition, @timholy suggested moving these to "operations.jl" as well. (cf. https://github.com/JuliaGraphics/ColorTypes.jl/pull/186#discussion_r411328795)

I haven't moved these because I don't see a clear reason.

kimikage:
I fixed a bug caused by `color_type(AGray) != Gray`, and added a few tests.

kimikage:
We have one more major task left to complete, so we will merge this first.
Which file to put the methods in is not the essential issue. It can be modified at any time.

cc: @timholy

PR #231, opened 2021-03-04
Version 0.11.0

kimikage:
#### TODO

- [x] Move `<` and gray-real comparison from ColorVectorSpace (PR #213)
- [x] Move comparison methods to "operations.jl" or something (PR #237)
  - cf. https://github.com/JuliaGraphics/ColorTypes.jl/pull/158#discussion_r376705144
- [x] Move `zero` for `{Abstract|Transparent}RGB` from ColorVectorSpace (PR #238)
- [x] Add `zero`/`oneunit` for more types (PR #238)
- [ ] Make a decision on the definition of `one` (PR #243) --> v0.12
  - cf. issue #235
- [x] Add error hints for color arithmetic (~a part of PR #243~ PR #249)
- [ ] Support `alpha(::Number)`? (PR #177) --> v0.12

Optional
- [x] Move `isfinite`, `isinf`, and `isnan` from ColorVectorSpace (PR #246)
  - They are not related to vector representations.
- [x] Move `nan` from ColorVectorSpace and add support for more types (PR #246)
  - cf. https://github.com/JuliaGraphics/ColorVectorSpace.jl/issues/75
- [x] Move `real(::Type{<:AbstractGray})` from ColorVectorSpace (PR #248)

#### Version table

I'm changing my mind on the policy of postponing PR #243 to ColorTypes v0.12 and releasing the corresponding ColorVectorSpace v0.10. I would like to proceed to ColorTypes v0.12.0 without releasing v0.11.x if possible.

|ColorTypes      | v0.10.10+| v0.11.0 | v0.12.0 |
|:---------------|:---------|:--------|:--------|
|Colors          |(v0.12.6+)| v0.12.8 | v0.13.0 |
|ColorVectorSpace|(v0.9.1+) | v0.9.4  |*v0.10.0*|

PR #243, opened 2021-04-03
[RFC] (Re-)define `one` for `Colorant`

### LINKED ISSUE #235: [RFC] Decision on the definition of `one`
We have been discussing the definition of `one` in PR #175 and #186 (as well as in some other places). However, I don't think the final decision has been made. I organize the candidates.

#### 1. Define `one(::Type{<:Colorant}) = 1`
If we take a color as a vector, it is reasonable to assume that the real number `1` is the identity for multiplication. Also, ColorTypes.jl has long announced `one(Gray) = 1` in depwarn (Most end users probably don't see depwarns anymore, though.).
We have not discussed anything other than grays and rgbs, but this idea can be generalized to `Colorant`, or at least should be generalized as a fallback.
The disadvantage of this idea is that `1` is not the "element" of a set of colors or color vectors. Of course, we have the `oneunit`.
Another problem is that `Int` is somewhat "strong" in the promotion rules. For now, we are choosing a policy of special treatment for integer arguments in color constructors (cf. #177, #197).

#### 2. Remove `one(::Type{<:Gray})` form ColorTypes.jl
As long as the definition of color multiplication is ambiguous and it is not defined in ColorTypes.jl, it is reasonable to not define `one`.
We have the error hinting now, which is better than the "invisible" depwarn.
Of course, this can lead to inconsistent definitions of `one`.

#### 3. Define `one(::Type{C}) where {C<:Colorant} = oneunit(C)`
This is reasonable if we define color multiplication (`*`) to be element-wise. Since the `*` operator is still "free" in ColorVectorSpace, there is room to add the definition.
The problem with this proposal is that it is different from the existing policy (at least the depwarn). Of course, neither ColorTypes.jl nor ColorVectorSpace.jl have reached v1 yet, so the policy change itself is not a problem.

cc: @timholy, @johnnychen94

kimikage:
This defines `one` as a real (i.e. dimensionless) number `１`.

I am not convinced that this is the best way, but I think this is a reasonable option.

Since this does not support `ones(ARGB, ...)` and so on, this may requires PR #177 ~as a part of v0.11.0~

Closes #235, #137

### DIFF
--- a/src/ColorTypes.jl
+++ b/src/ColorTypes.jl
@@ -6,7 +6,7 @@ using Base: @pure
 
 const Fractional = Union{AbstractFloat, FixedPoint}
 
-import Base: ==, <, isless, isapprox, isfinite, isinf, isnan, oneunit, zero,
+import Base: ==, <, isless, isapprox, isfinite, isinf, isnan, one, oneunit, zero,
              hash, eltype, length, real, convert, reinterpret, show
 using Random
 import Random: rand
--- a/src/error_hints.jl
+++ b/src/error_hints.jl
@@ -2,22 +2,15 @@ MathTypes{T,C} = Union{AbstractRGB{T},TransparentRGB{C,T},AbstractGray{T},Transp
 
 # provided by https://github.com/JuliaLang/julia/pull/35094
 function register_hints()
-    if isdefined(Base, :Experimental) && isdefined(Base.Experimental, :register_error_hint)
-        Base.Experimental.register_error_hint(MethodError) do io, exc, argtypes, kwargs
-            if exc.f in (one,) && argtypes[1] <: Union{Type{<:AbstractRGB}, AbstractRGB}
-                print(io, "\nYou may need to `using ColorVectorSpace`.")
-            end
-            if exc.f in (ones,) && argtypes[1] <: Type{<:AbstractRGB}
-                print(io, "\nYou may need to `using ColorVectorSpace`.")
-            end
-            # In theory we could list every function supported by ColorVectorSpace.
-            # This list of functions is far from comprehensive but may be enough to catch many users.
-            if exc.f in (+, -, *, /) && any(T->T<:MathTypes || T<:Type{<:MathTypes}, argtypes)
-                print(io, "\nMath on colors is deliberately undefined in ColorTypes, but see the ColorVectorSpace package.")
-            end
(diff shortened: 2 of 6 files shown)


johnnychen94 (approved):
No objections at all!

Remind me, what's the plan on ColorVectorSpaces? Will there be any `zero`/`one`/`oneunit` methods there? I guess no?

kimikage:
I'm not sure I fully understand what @timholy wants to do with ColorVectorSpace v0.9. For now, ColorVectorSpace is allowing breaking changes to remain at v0.9, rather than "skipping" to ColorVectorSpace v0.10. (Incidentally, I'd rather there be no gaps than to strictly adhere to the SemVer principles.)

The options can be roughly divided into two approaches.

One is that ColorTypes v0.11 will not include this PR, and only define `zero` and `oneunit` (and `<`). This is a "formally" safe approach, but "in practice" it may cause confusion about the support for ColorVectorSpace v0.9 in the downstream packages.

The other is to include this PR and PR #177 in ColorTypes v0.11 and break the definition of `one` in ColorVectorSpace v0.9.x, and drop the support for ColorTypes v0.10. This seems to be a hard landing, but the package manager should avoid the crash.:sweat_smile:

kimikage:
In the meantime, I will make some trivial PRs for ColorVectorSpace.

kimikage:
ImageCore v0.9.0, which depends on ColorVectorSpace v0.9, has been released. Currently it should not be available yet, but some packages will start supporting ColorVectorSpace v0.9. Therefore, releasing ColorVectorSpace v0.10 may be the right way to go.

kimikage:
I've updated the plan: #231.  I'll merge PR #243 and another migration PR, and release v0.11.0 by the beginning of next week. v0.12.0 is a bit heavier, but I hope to release it by the end of May. I think FixedPointNumbers v0.9 will be the next target.
Of course, this is a personal observation and not a definitive one, for better or worse.

kimikage:
Ah, the error hints for color arithmetic may be v0.11 material. I'm going to split the PR further.

johnnychen94:
Since error hints don't change the behavior so it's okay to ship it in v0.12 just to save some efforts; "whoever wants the new fancy feature, use the new version."

kimikage:
I have submitted a separated PR #249. I am going to remove duplicate changes when I apply `rebase` to this PR. That is, I will leave this PR as is for a while.

PR #245, opened 2021-04-08
Improve `compN` for non-parametric `ColorAlpha` types

kimikage:
Since only `AGray32` and `ARGB32` are defined as non-parametric `TransparentColor` types, I did not notice this problem, but it would be better to specialize `compN` for `ColorAlpha` as well as `AlphaColor`.

This is not an urgent issue, since it can be solved by overloading `compN` by users or packages which define custom color types. Therefore, I will make this a part of v0.12 instead of v0.11, even though this is a non-breaking change as long as `alpha(c)` works.


kimikage:
It's been half a month since the release of v0.11.0, so I'd like to start the v0.12 series.

PR #250, opened 2021-05-05
[RFC/WIP] Update README.md

kimikage:
This is a proposed update based on the changes planned for ColorTypes v0.12. (see PR #251)
I think the update should be merged at the final stage of the releasing, since the documentation for ColorTypes (i.e. README.md) is always published in the development. On the other hand, we can apply the parts which are not related to the changes in v0.12 successively.

The important thing is to clarify the default colorspace for `YIQ` and `YCbCr`.

xref: https://github.com/JuliaGraphics/Colors.jl/issues/453
xref: https://discourse.julialang.org/t/what-should-the-default-color-space-for-ycbcr-be/60874

Closes #10, Closes #30

PR #252, opened 2021-05-07
Make `AbstractGray` an abstract type instead of an alias

### LINKED ISSUE #230: [RFC] Distinguishing between `AbstractGray{T}` and `Color{T, 1}`
Currently, `AbstractGray{T}` is the alias for `Color{T, 1}`, but this is a bit weird. (cf. https://github.com/JuliaGraphics/ColorTypes.jl/issues/145#issuecomment-573071205)

Here is an extreme example.

```julia
struct Lightness{T} <: Color{T, 1}
     v::T
end

struct Darkness{T} <: Color{T, 1}
     v::T
end

ColorTypes.gray(c::Lightness) = c.v
ColorTypes.gray(c::Darkness) = oneunit(c.v) - c.v
ColorTypes.comp1(c::Darkness) = c.v
```

```julia
julia> Darkness(0.2) == Lightness(0.8) # OK
true

julia> Darkness(0.2) > Lightness(0.3) # Uh, yes, that's correct.
true

julia> Darkness(0.2) == 0.8 # What? Oh, hmm.
true 
```
Colored tones such as sepia and cyanotype can be prevented from strange reinterpretations by not implementing `gray()`. However, when you implement `gray()` for a "gray" scale, strange things happen.

That is, color types which have difference between `comp1` and `gray` are inherently incompatible with `AbstractGray`.

IMO, `AbstractGray` should be an `abstract type` as well as `AbsractRGB`.

kimikage:
This classifies the properties of real numbers and the properties of the black-gray-white colors.

This should be mostly compatible for a single specific method, but may affect the multiple dispatch or ambiguity.

README.md will be updated with PR #250.

Closes #230

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -16,6 +16,17 @@ grayscale) with no transparency.
 abstract type Color{T, N} <: Colorant{T,N} end
 
 """
+`AbstractGray{T}` is an abstract supertype for gray types which corresponds to
+real numbers, where `0` means black and `1` means white.
+
+!!! compat "ColorTypes v0.12"
+    Prior to ColorTypes v0.12, `AbstractGray{T}` was the alias for `Color{T,1}`;
+    after ColorTypes v0.12, `Color{T,1}` does not necessarily correspond to the
+    black-gray-white colors.
+"""
+abstract type AbstractGray{T} <: Color{T,1} end
+
+"""
 `AbstractRGB{T}` is an abstract supertype for red/green/blue color types that
 can be constructed as `C(r, g, b)` and for which the elements can be
 extracted as `red(c)`, `green(c)`, `blue(c)`. You should *not* make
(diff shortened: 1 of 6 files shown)


src/types.jl
```diff
@@ -15,6 +15,17 @@ grayscale) with no transparency.
 """
 abstract type Color{T, N} <: Colorant{T,N} end
 
+"""
+`AbstractGray{T}` is an abstract supertype for gray types which corresponds to
+real numbers, where `0` means black and `1` means white.
+
+!!! compat "ColorTypes v0.12"
+    Prior to ColorTypes v0.12, `AbstractGray{T}` was the alias for `Color{T,1}`;
+    after ColorTypes v0.12, `Color{T,1}`` does not necessarily correspond to the
```
johnnychen94, line 24:
```suggestion
    after ColorTypes v0.12, `Color{T,1}` does not necessarily correspond to the
```

PR #254, opened 2021-05-15
Update `gamutmin`/`gamutmax` based on sRGB gamut

### LINKED ISSUE #125: `gamutmax(LCHab)` disagrees with the Colors.jl usage

kimikage:
Fixes #125

### DIFF
--- a/src/operations.jl
+++ b/src/operations.jl
@@ -44,21 +44,44 @@ hash(c::TransparentRGB, hx::UInt) = hash(alpha(c), hash(blue(c), hash(green(c), 
 Base.adjoint(c::Colorant) = c
 
 # gamut{min,max}
-gamutmax(::Type{T}) where {T<:HSV} = (360,1,1)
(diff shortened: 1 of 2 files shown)


kimikage:
If you notice any inappropriate values, please report them. Perhaps the values for `YCbCr` needs to be redefined.

PR #256, opened 2021-05-27
Modify error messages in constructors

### LINKED ISSUE #242: [RFC] Warn about the normalization first in error messages of RGB constructors
I have recently seen a misuse like `RGB(255, 0.0, 0.0)`.
It is inconceivable that such a misuse could happen without any context. If I were to speculate on the user's actions, the user would have written something like `RGB(255, 0, 0)` first. Then, the following error message should have been displayed.
```julia
julia> RGB(255, 0, 0)
ERROR: ArgumentError: (255, 0, 0) are integers in the range 0-255, but integer inputs are encoded with the N0f8
  type, an 8-bit type representing 256 discrete values between 0 and 1.
  Consider dividing your input values by 255, for example: RGB{N0f8}(255/255,0/255,0/255)
  Or use `reinterpret(N0f8, x)` if `x` is a `UInt8`.
```
I think that error message is more than kind enough. However, maybe some users will stop reading in the middle of the first sentence.
The first sentence indirectly mentions the normalization, but it can also be interpreted as a restriction on the specific encoding of `N0f8`. Then, there may be an incentive to use types other than `N0f8` or integers. The trouble is, that is not exactly wrong.

I think it would be better to first mention the normalization as a common rule of `AbstractRGB`, then give an example of `RGB{N0f8}(255/255, 0/255, 0/255)`, and finally explain `N0f8`.

It might be a good idea to color the example to make it stand out.

kimikage:
Closes #242

```
julia> RGB(255, 0, 0) # This PR
ERROR: ArgumentError: The components of RGB are normalized to the range 0-1,
  but (255, 0, 0) are integers in the range 0-255.
  Consider dividing your input values by 255, for example: RGB{N0f8}(255/255, 0/255, 0/255)
  The component type N0f8 is an 8-bit type representing 256 values from 0 to 1.
  You can also use `reinterpret(N0f8, x % UInt8)` to encode the input into N0f8.
```

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -117,13 +117,6 @@ convert(::Type{T}, x::Gray24) where {T<:Real} = convert(T, gray(x))
 real(x::ColorantN{1}) = comp1(x)
 real(x::Type{<:ColorantN{1}}) = real(eltype(x))
 
-# Define some constructors that just call convert since the fallback constructor in Base
-# is removed in Julia 0.7
-# The parametric types are handled in @make_constructors and @make_alpha
-for t in (:ARGB32, :Gray24, :RGB24)
-    @eval $t(x) = convert($t, x)
-end
-
 # reinterpret
 for T in (RGB24, ARGB32, Gray24, AGray32)
     @eval begin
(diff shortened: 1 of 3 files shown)


kimikage:
FYI, as for https://github.com/JuliaGraphics/ColorTypes.jl/pull/187#discussion_r445370098, this PR and #255 alone are not a complete solution. (Currently, it is degraded.)
Adding generic constructors for `AbstractRGB` etc in PR #197 will complete it.

Issue #247, opened 2021-04-22
`init` keyword for `reducec` and `mapreducec`


johnnychen94:
The current API is not consistent with `reduce`/`mapreduce`.

```diff
- reducec(op, v0, c)
+ reducec(op, c; init=zero(eltype(eltype(c))))

- mapreducec(f, op, v0, c)
+ mapreducec(f, op, c; init=zero(eltype(eltype(c)))
```

Maybe, we can just make `mapc`/`reducec`/`mapreducec` as Colorant methods of `map`/`reduce`/`mapreduce`?

kimikage:
I agree with the support of the keyword argument `init`. However, there is room for debate about compatibility.
If we support multiple input colors, (as the current `mapc` supports,) there will be some confusion. For example:
```julia
mapreducec((a, b) -> max(a, b), *, Gray(1), Gray(0)) # Does `Gray(1)` mean the `v0`?
```
I don't think this is much of a problem in the current situation. However, issue #207 needs to be considered more carefully.
If the current formats are to be deprecated, we need to prepare for `depwarn`.

BTW, this isn't an essential issue, but the default value of `init` seems to depend on the operator `op`.
```julia
julia> mapreduce(identity, +, (1.0, 0.5, 0.25))
1.75

julia> mapreduce(identity, +, (1.0, 0.5, 0.25), init=0)
1.75

julia> mapreduce(identity, *, (1.0, 0.5, 0.25))
0.125

julia> mapreduce(identity, *, (1.0, 0.5, 0.25), init=1)
0.125
```

kimikage:
I also set one of the milestones for v0.13 to be the provision of iterator interfaces (cf. https://github.com/JuliaGraphics/ColorTypes.jl/issues/131#issuecomment-824544538)
In that case, `reducec` and `mapreducec` may not be necessary anymore.
```julia
julia> components(c::Color3) = (comp1(c), comp2(c), comp3(c)); # substitute a tuple for an iterator object

julia> mapreduce(identity, *, components(RGB(1.0, 0.5, 0.25)))
0.125

julia> mapreducec(identity, *, 1.0, RGB(1.0, 0.5, 0.25))
0.125

```

kimikage:
Even if we may deprecate `reducec` and `mapreducec` in the future, I think it would be beneficial to change their implementation to `reduce`/`mapreduce`-based (unless there is a significant performance degradation).

[...]

PR #260, opened 2021-07-31
Add component iterator

kimikage:
This adds a component iterator as a "private" API.
This re-implements the folding functions with the iterator.

cf. #190, #247

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -126,3 +126,6 @@ coloralpha(c::C) where {C<:Color} = coloralpha(C)(c)
 coloralpha(c::C,a) where {C<:Color} = coloralpha(C)(c,a)
 coloralpha(c::C) where {C<:TransparentColor} = coloralpha(base_color_type(C))(color(c), alpha(c))
 coloralpha(c::C,a) where {C<:TransparentColor} = coloralpha(base_color_type(C))(color(c), a)
+
+# Tuple
+Tuple(c::Colorant{T, N}) where {T, N} = (comps(c)...,)::NTuple{N, T}
--- a/src/operations.jl
+++ b/src/operations.jl
@@ -1,5 +1,43 @@
-struct BoolTuple end
-@inline BoolTuple(args::Bool...) = args
+# iterator
+struct ComponentIterator{C<:Colorant}
+    c::C
+end
+
+eltype(::Type{ComponentIterator{C}}) where {T, C <: Colorant{T}} = T
+length(::ComponentIterator{C}) where {N, C <: ColorantN{N}} = N
+
+function Base.iterate(itr::ComponentIterator{C}, state::Int=0) where {N, C <: ColorantN{N}}
+    state < 0 && return nothing
+    state >= N && return nothing
+    state += 1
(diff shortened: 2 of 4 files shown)


src/operations.jl
```diff
@@ -1,5 +1,34 @@
...
+    N > 1 && i == 2 && return comp2(itr.c)
+    N > 2 && i == 3 && return comp3(itr.c)
+    N > 3 && i == 4 && return comp4(itr.c)
+    N > 4 && i == 5 && return comp5(itr.c)
+    throw(BoundsError(itr, i))
+end
+Base.axes(::ComponentIterator{C}) where {N, C <: ColorantN{N}} = (Base.OneTo(N),)
+Base.ndims(::Type{ComponentIterator{C}}) where {C} = 1
+function Base.broadcastable(itr::ComponentIterator{C}) where {T, N, C <: Colorant{T, N}}
+    (itr...,)::NTuple{N, T}
```
kimikage, line 27: I think this should return the iterator itself (`itr`) instead of the tuple, but I don't know how to handle that well. :confused:

src/conversions.jl
```diff
@@ -126,3 +126,6 @@ coloralpha(c::C) where {C<:Color} = coloralpha(C)(c)
 coloralpha(c::C,a) where {C<:Color} = coloralpha(C)(c,a)
 coloralpha(c::C) where {C<:TransparentColor} = coloralpha(base_color_type(C))(color(c), alpha(c))
 coloralpha(c::C,a) where {C<:TransparentColor} = coloralpha(base_color_type(C))(color(c), a)
+
+# Tuple
+Tuple(c::Colorant{T, N}) where {T, N} = (comps(c)...,)::NTuple{N, T}
```
johnnychen94, line 131: Is this the recommended API that people should use?

kimikage: Strictly speaking, I don't think the `Tuple` constructors are a part of public APIs because they don't have docstrings.
However, I don't see a problem with ColorVectorSpace.jl etc. using it internally.

kimikage: Of course, some people might argue that `Colorant` is never iterable. However, since the design of `tuple()` and `Tuple()` are clearly different, I don't think there will be any confusion.

kimikage: I think we need to discuss again whether `Tuple` should be overloaded, and if so, what should be returned. (cf. PR #184)
Currently (ColorTypes v0.11) `Tuple(::Colorant)` causes an error.
```julia
julia> Tuple(RGB(1.0, 0.0, 0.0))
ERROR: MethodError: no method matching iterate(::RGB{Float64})
```

johnnychen94: I'm personally in favor of the `Tuple(::Colorant)` usage. It's kind of like `CartesianIndex`, that you can't iterate it but can easily construct an iterable tuple from `Tuple` or `x.I`.

kimikage:
There is still a lot of room for discussion and improvement, but the iteration feature is useful for addressing the rest of the issues, so I will merge this first.

Issue #207, opened 2020-07-09
let `mapc` support Number types


johnnychen94:
```julia
julia> mapc(*, x, y)
ERROR: MethodError: no method matching _same_colorspace(::Float64, ::Float64)
Stacktrace:
 [1] mapc(::typeof(*), ::Float64, ::Float64) at /Users/jc/.julia/packages/ColorTypes/GHtY7/src/operations.jl:107
 [2] top-level scope at REPL[4]:1

julia> mapc(*, Gray(x), Gray(y))
Gray{Float64}(0.013853760001569756)
```

I believe this helps unify how we process `AbstractArray{<:AbstractGray}` and `AbstractArray{<:Number}`.

I can put my hands on it if you think this is a good idea.

kimikage:
There is room for consideration about the return type.
Perhaps we need to determine the promotion rules for `AbstractGray` and `Number`.  I also think the promotion should be consistent with PR #177.

johnnychen94:
> Perhaps we need to determine the promotion rules for AbstractGray and Number.

If we don't allow `mapc(op, x::Union{AbstractGray, Number}, y::Union{AbstractGray, Number}`, I think it would just be fine as the return type of `mapc(op, ::Number, ::Number)` will be defined by `op`.

kimikage:
```julia
julia> mapc(*, Gray( Int(1) ), Gray( Bool(0) ))
Gray{N0f8}(0.0)

julia> mapc(*, Int(1), Bool(0))
0

julia> mapc(*, Complex(1), Rational(0))
0//1 + 0//1*im
```
:ghost:

PR #263, opened 2021-08-13
Add `init` keyword support for `reducec` & `mapreducec`

### LINKED ISSUE #247: `init` keyword for `reducec` and `mapreducec`

kimikage:
This also makes it possible to omit specifying the `init` value.

```julia
julia> rgb = rand(RGB, 1000, 1000);

julia> @btime reducec.(+, 0.2, rgb); # conventional
  1.746 ms (6 allocations: 7.63 MiB)

julia> @btime reducec.(+, rgb, init=0.2); # with `init`
  1.897 ms (13 allocations: 7.63 MiB)

julia> @btime reducec.(+, rgb); # omitting `init` (=`0.0`)
  1.883 ms (5 allocations: 7.63 MiB)
```

Closes #247

### DIFF
--- a/src/operations.jl
+++ b/src/operations.jl
@@ -243,10 +243,11 @@ _same_colorspace(::Type{C}, ::Type{C}) where {C<:Colorant} = C
 
 """
     reducec(op, v0, c)
+    reducec(op, c; [init])
 
-Reduce across color channels of `c` with the binary operator
-`op`. `v0` is the neutral element used to initiate the reduction. For
-grayscale,
+Reduce across color channels of `c` with the binary operator `op`.
+`v0` or `init` is the neutral element used to initiate the reduction.
+For grayscale,
 
     reducec(op, v0, c::Gray) = op(v0, comp1(c))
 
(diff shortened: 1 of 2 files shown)


timholy (approved):

kimikage:
After sorting out the handling of `Number` input, let's decide if we want to support multiple iterators. (cf. #207, https://github.com/JuliaGraphics/ColorTypes.jl/issues/247#issuecomment-824607503)
The conventional style (`v0`) is also available for now.

johnnychen94:
I'm very much in favor of `mapc`/`mapreducec` with multiple iterators, it makes color types more like iterator without breaking their atomic/non-broadcastable property.

PR #272, opened 2022-05-10
Support `isequal`

timholy:
Formerly we fell back to `==`, but this does not do the right thing
for `NaN` components.


johnnychen94 (approved):

timholy:
Let's run CI (I re-enabled UnitTest).

PR #278, opened 2022-05-15
add missing Gray{T} constructor for colorant

### LINKED ISSUE #277: `Gray{T}(c::RGB)` doesn't work anymore
```julia
(@v1.7) pkg> activate --temp
  Activating new project at `/tmp/jl_u6D0ay`

(jl_u6D0ay) pkg> add ColorTypes@0.11.1 Colors
...

julia> using Colors

julia> Gray{Float32}(rand(RGB))
ERROR: MethodError: Cannot `convert` an object of type RGB{Float64} to an object of type Float32
Closest candidates are:
  convert(::Type{T}, ::T) where T<:Number at ~/packages/julias/julia-1.7/share/julia/base/number.jl:6
  convert(::Type{T}, ::Number) where T<:Number at ~/packages/julias/julia-1.7/share/julia/base/number.jl:7
  convert(::Type{T}, ::Base.TwicePrecision) where T<:Number at ~/packages/julias/julia-1.7/share/julia/base/twiceprecision.jl:262
  ...
Stacktrace:
 [1] Gray{Float32}(val::RGB{Float64})
   @ ColorTypes ~/.julia/packages/ColorTypes/LWFPo/src/types.jl:345
 [2] top-level scope
   @ REPL[7]:1
```

This works for 0.11.0

johnnychen94:
fixes #277

### DIFF
--- a/src/types.jl
+++ b/src/types.jl
@@ -515,6 +515,7 @@ function (::Type{C})() where {N, C <: ColorantN{N}}
     C(ntuple(_ -> d0, Val(N - 1))..., dx)
 end
 
+Gray{T}(c::Colorant) where {T<:Union{Fractional,Bool}} = _new_colorant(Gray{T}, c)
 (::Type{C})(x         ) where {C <: Color    } = _new_colorant(C, x)
 (::Type{C})(x, y      ) where {C <: ColorN{2}} = _new_colorant(C, x, y)
 (::Type{C})(x, y, z   ) where {C <: ColorN{3}} = _new_colorant(C, x, y, z)
--- a/test/conversions.jl
+++ b/test/conversions.jl
@@ -753,6 +753,14 @@ end
     end
 end
 
+@testset "conversions from rgb to gray" begin
+    # The rules are defined in Colors.jl, but we want to make sure it calls
+    # the `convert` method.
+    # Issue #277
+    msg = "No conversion of RGB{N0f8}(0.0, 0.0, 0.0) to Gray{Float32} has been defined"
+    @test_throws ErrorException(msg) Gray{Float32}(RGB())
+end
+
 @testset "conversions between different spaces" begin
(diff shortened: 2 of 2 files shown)


src/types.jl
```diff
@@ -515,6 +515,7 @@ function (::Type{C})() where {N, C <: ColorantN{N}}
     C(ntuple(_ -> d0, Val(N - 1))..., dx)
 end
 
+Gray{T}(c::Colorant) where {T<:Union{Fractional,Bool}} = _new_colorant(Gray{T}, c)
```
johnnychen94, line 518: Unsure why, I tried the following but it doesn't work.

```julia
(::Type{C})(x ) where {C <: Gray } = _new_colorant(C, x)
```

kimikage:
@johnnychen94 I apologize for leaving the development unattended. Thanks for your maintenance, it is much appreciated.

By the way, why is the constructor of `Gray{T}` needed? It is the job of Colors.jl to convert `RGB` to `Gray`.
And I wanted to enhance support for `AbstractGray` other than `Gray` a couple of years ago.

johnnychen94:
>  It is the job of Colors.jl to convert RGB to Gray.

I wasn't that clear on this idea until you mentioned this. The intention was to fix some compatibility issues during the transition from 0.11.0 to 0.11.1 (which should be backward compatible).

During the time @kimikage you were unavailable, nobody realized that the current master contains some unexpected breaking changes, and [we (actually, Tim) just released them as patch releases](https://github.com/JuliaGraphics/ColorTypes.jl/commit/b8ed378531f1445ed2402a00d49ecbf64ce5a0fc), which was incorrect. This PR is a fixup to that.

So, if you decide to move the conversions to Colors, you can safely revert this patch commit in the master branch and continue the development of 0.12 until you're ready.

Issue #273, opened 2022-05-15
ColorTypes v0.11.1 is perhaps too breaking


johnnychen94:
This new version breaks the test for ImageCore, ImageIO, ImageQualityIndexes, and perhaps more...

I'm currently occupied with other stuff so don't have time to find out what's really breaking, so how about reverting all those changes, tag a new v0.11.2 release to override the breaking v0.11.1 version.
Then we can tag a 0.12 release so that we have more time to gradually upgrade it?

issue list:

- #275 
- #277 
- https://github.com/JuliaImages/ImageCore.jl/pull/181

johnnychen94:
update: #276 and #278 fixes the test in ImageIO and ImageQualityIndexes

timholy:
Copied from elsewhere: 

Oh shoot. I didn't check how much was on master, big mistake.

I will try to fix this. I think the best plan is to fork a release 0.11 branch, backport my recent commit, and release 0.11.2. Reasonable?

timholy:
Fair to say this was closed by #280?

PR #281, opened 2022-05-15
ColorTypes v0.12.0-dev

johnnychen94:
To avoid unexpected version bump such as #273

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ColorTypes"
 uuid = "3da002f7-5984-5a60-b8a6-cbb66c0b333f"
-version = "0.11.1"
+version = "0.12.0-dev"
 
 [deps]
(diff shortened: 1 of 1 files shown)


kimikage:
@johnnychen94
I believe that this is definitely better to have because an "accident" have actually happened.

However, there is some ambiguity regarding the intent of this.
That is, I don't know if these “-DEV” declarations are restrictions on the package itself, or restrictions that also apply to downstream packages.
To change the perspective, the question is whether ColorTypes v0.12.0 should include FixedPointNumbers v0.9.0-DEV (or simply v0.9) in the `[compat]`, or not.

As of three years ago, ColorTypes v0.12 was scheduled to drop support for FixedPointNumbers v0.8 and earlier.
While some planning changes are necessary, I believe that is still a reasonable choice as it stands.
Dropping can be done at release time, but adding must be done during development.

PR #282, opened 2022-05-24
Implement 3-colorant `mapc`

timholy:
This will be useful for supporting `clamp`


johnnychen94 (approved):
Needs to update the docstring R139

```
"""
"""
    mapc(f, rgb) -> rgbf
    mapc(f, rgb1, rgb2) -> rgbf
...
"""
```

kimikage:
It would probably be kinder to have a compat admonition.

PR #286, opened 2022-11-23
Preemptively fix tests that are about to break

apaz-cli:
Apply these changes if https://github.com/JuliaLang/julia/pull/47369 gets merged.

The error message for MethodError currently applies to both calls to functions and `invoke()`, but did not specify which the error came from. Fixing this will break some tests, hence this PR.

### DIFF
--- a/test/error_hints.jl
+++ b/test/error_hints.jl
@@ -19,11 +19,11 @@ end
         gray = Gray(0.8)
         rgb = RGB{Float32}(1, 0, 0)
         err_str = @except_str gray + rgb MethodError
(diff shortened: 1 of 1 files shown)


kimikage:
Please make sure to keep the tests robust, as this package will also be compatible with the older julia.

apaz-cli:
The PR mentioned was merged last February. If the tests aren't failing for newer versions, you can close the PR.

kimikage:
@apaz-cli Wouldn't it be better to revisit the PRs in other packages as well?

PR #290, opened 2023-07-07
remove tests that check that Base does not have certain features

KristofferC:
Tests like this that check that certain features in Base do not exist are kind of pointless. They start failing once Base implements does features and the only thing to do is to remove the tests. In 1.10 this does not error so the tests fail.


KristofferC:
cc @timholy

KristofferC:
cc @timholy

timholy:
Sorry for the delay. Doing a release is a bit complicated because we have breaking stuff merged to master. Do you need me to backport this? (Is this about PkgEval?)

KristofferC:
No rush, I know I've reported this here so I can filter it out.

PR #271, opened 2022-01-01
Add Oklab perceptual color space

nealmckee:
Implements the Oklab color space as described here: https://bottosson.github.io/posts/oklab/

Adds and exports the type.
Adds Oklab and LCHOklab to the chroma and hue trait accessor functions.
Adds description to the readme.

This is my first contribution to Julia, and any software package in general so help with any oversights would be greatly appreciated. (Am I missing tests for example? They weren’t mentioned in the guidelines of how to extend color types.)

PR #251, opened 2021-05-06
Version 0.12.0

kimikage:
#### TODO

- [x] Improve `compN` for non-parametric `ColorAlpha` types (PR #245)
- [x] Simplify display of colors (PR #206)
  - This can be a cause of failures in the testing of downstream packages, but I would rather merge it early to have time to fix the tests.
- [x] (Re-)define `one` for `Colorant` (PR #243)
- [x] Support `alpha(::Number)` (PR #177)
- [x] Use sampler-based Random API (PR #222)
- [x] Update `gamutmin`/`gamutmax` based on sRGB gamut (PR #254)
  - cf. issue #125
- [x] Make `AbstractGray` an abstract type instead of an alias (PR #252)
  - cf. issue #230
- [x] Generalize conversions from/to `Real` (PR #255)
- [x] Remove `eltype_ub` (PR #258)
  - cf. issue #257 
- [x] Redesign color constructors (PR #197)
- [x] Generalize conversions within the same base color
  - cf. issue #241
- [ ] Fix conversions from parametric colors to abstract colors
  - cf. `@test_broken`s in "test/conversions.jl"
- [x] Modify error messages in constructors (PR #256)
  - cf. issue #242
  - cf. https://github.com/JuliaGraphics/ColorTypes.jl/pull/187#discussion_r445370098
- [x] Add a "function" which returns a simple component iterator (PR #260)
  - cf. issue #190
  - In v0.12, it may be a private API. The enhancements will be made in ColorTypes v0.13.
- [x] Re-implement folding functions (e.g. `reducec`) with the component iterator (PR #263)
  - cf. issue #247
- [ ] Decide about the deprecation of `eltype` and `length`
  - cf. PR #184
- [ ] Define a new API to replace `eltype_default`
  - cf. issue #257
- [ ] Optimize `hue(c)` (PR #262)
  - cf. https://github.com/JuliaGraphics/Colors.jl/pull/495
- [ ] Update README.md (PR #250)
  - PR #250 will be separated into a number of PRs.

#### Version table

|ColorTypes       | v0.11.0  | v0.12.0    | v0.13.0 |
|:----------------|:---------|:-----------|:--------|
|FixedPointNumbers|(v0.5-0.8)| v0.6.1-0.8 | v0.9+   |
|Colors           |(v0.12.8) | v0.13.0    | v0.13.x?|
|ColorVectorSpace |(v0.9.4+) | v0.10.0    | v0.10.x?|
[...]

PR #291, opened 2023-08-01
Add Oklab and Oklch

eprovst:
Similar to pull request #271 using the terminology of CSS instead and including some tests.

As far as I can tell, I implemented everything needed. The `gamutmin` and `gamutmax` are from the CSS recommendation, but I'm not 100% certain it stays within the sRGB gamut, rather likely it does not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,6 +252,25 @@ struct LCHuv{T} <: Color{T,3}
 end
 ```
 
+### Oklab and Oklch
+
+A perceptually uniform colorspace developed by
+[Björn Ottosson](https://bottosson.github.io/posts/oklab/) and its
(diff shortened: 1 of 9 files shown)


timholy (approved):
LGTM. OK to merge?

timholy:
That's as good as some of the other colorspaces, so I'm fine with that.

timholy:
Thanks!

Getting this out in a release is a little complicated: we already have some breaking stuff merged and there's still #251. Are you in a hurry to get this out or is a bit of a delay OK?

eprovst:
It would be useful for some long term project, but there's no rush. Of course, the complementary pull request in Colors.jl depends on a release here.

PR #301, opened 2024-04-25
Update workflows

kimikage:
This adds unit tests for julia v1.6 and julia on aarch64 (with QEMU).
This also introduces Dependabot.


kimikage:
Since the workflows are basically from the default branch (i.e. `master`), there is no need to barkport this to v0.11.
However, this is worth backporting as it allows for the `workflow_dispatch` triggers.

PR #302, opened 2024-04-25
Add test and followup for `StyledStringsExt`

kimikage:
I do not understand the need for me to create this PR, but this PR itself might be necessary for sound development.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -14,12 +14,13 @@ StyledStringsExt = "StyledStrings"
 
 [compat]
 FixedPointNumbers = "0.5, 0.6, 0.7, 0.8"
-StyledStrings = "1"
+# StyledStrings = "1" # FIXME: workaround for unregistered StyledStrings
 julia = "1"
 
 [extras]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f" # FIXME: workaround for unregistered StyledStrings
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Documenter", "Test"]
+test = ["Documenter", "Pkg", "Test"]
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,3 +1,14 @@
+
+# FIXME: Once StyledStrings v1 is registered, remove the following and the `Pkg` dependency
+if isdefined(Base, :get_extension)
+    import Pkg
+    if VERSION < v"1.11"
+        Pkg.add(url="https://github.com/JuliaLang/StyledStrings.jl", rev="julia1-compat")
+    else
+        Pkg.add("StyledStrings")
+    end
+end
+
 using ColorTypes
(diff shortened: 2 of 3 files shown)


kimikage:
@tecosaur Is the following intended behavior?
```julia
julia> StyledStrings.SimpleColor(RGB(1, 0, 0))
ERROR: MethodError: Cannot `convert` an object of type
  RGB{FixedPointNumbers.N0f8} to an object of type
  Union{@NamedTuple{r::UInt8, g::UInt8, b::UInt8}, Symbol}
The function `convert` exists, but no method is defined for this combination of argument types.
```

Project.toml
```diff
@@ -14,12 +14,13 @@ StyledStringsExt = "StyledStrings"
 
 [compat]
 FixedPointNumbers = "0.5, 0.6, 0.7, 0.8"
-StyledStrings = "1"
+# StyledStrings = "1" # FIXME: workaround for unregistered StyledStrings
```
kimikage, line 17: This should be fixed in the close future, but currently `StyledStrings` is unregistered and it should be v1 for the time being.
We are touching the internals of `StyledStrings`, so we run the risk of incompatibility. And that can't be prevented by the `"1"` cap anyway.
（In other words, *if we follow the principle*, we need to specify the patch version here.）

test/runtests.jl
```diff
@@ -1,3 +1,14 @@
+
+# FIXME: Once StyledStrings v1 is registered, remove the following and the `Pkg` dependency
+if isdefined(Base, :get_extension)
+    import Pkg
+    if VERSION < v"1.11"
+        Pkg.add(url="https://github.com/JuliaLang/StyledStrings.jl", rev="julia1-compat")
```
kimikage, line 6: I don't know how long `julia1-compat` remains in effect, but it is more practical than `main`.

tecosaur: I don't anticipate deleting the branch at any point, it will just become abandoned at some stage.

tecosaur:
> @tecosaur Is the following intended behavior?
> ```julia
> julia> StyledStrings.SimpleColor(RGB(1, 0, 0))
> ERROR: MethodError: Cannot `convert` an object of type
>   RGB{FixedPointNumbers.N0f8} to an object of type
>   Union{@NamedTuple{r::UInt8, g::UInt8, b::UInt8}, Symbol}
> The function `convert` exists, but no method is defined for this combination of argument types.
> ```

I'd say so, given I didn't intend for the constructor to work directly with `Colorant`s regardless of the conversion methods implemented.

kimikage:
> I didn't intend for the constructor to work directly with `Colorant`s

If you do not plan to make any changes in this package, I am ok with it.
Now let's merge this.

Project.toml
kimikage: What constitutes a breaking change is ambiguous. If `SimpleColor` added a feature to have an alpha component, it would not break the code, but it would go against the user's intuition.
That is just an example. `StyledStrings` is community owned, although you are the lead developer. I don't believe anything bad will happen, but there is no guarantee how it will turn out.
*For example, CI could cease to function effectively even though no one has done anything wrong.*

Issue #288, opened 2022-12-14
ARGB32 (and/or AGray32) test failure on aarch64


Millak:
I'm attempting to package ColorTypes.jl in Guix and I came across a test failure on aarch64. I have 25 instances of:

> RGB24/ARGB32 constructors: Test Failed at /gnu/store/ms5wncdgzv1p90xahj6pk6xk8lfx4c6a-julia-colortypes-0.11.1/share/julia/loadpath/ColorTypes/test/types.jl:177
>  Expression: ARGB32(c0, 0.8) === ARGB32(val1, val2, val1, 0.8)
>   Evaluated: ARGB32(0.2N0f8, 0.6N0f8, 0.2N0f8, 0.0N0f8) === ARGB32(0.2N0f8, 0.6N0f8, 0.2N0f8, 0.8N0f8)
>Stacktrace:
> [1] macro expansion
>   @ /gnu/store/q92hjagi3kxzqajvx1qp1jz7az747y98-julia-1.8.3/share/julia/stdlib/v1.8/Test/src/Test.jl:464 [inlined]
> [2] macro expansion
>   @ /gnu/store/ms5wncdgzv1p90xahj6pk6xk8lfx4c6a-julia-colortypes-0.11.1/share/julia/loadpath/ColorTypes/test/types.jl:177 [inlined]
> [3] macro expansion
>   @ /gnu/store/q92hjagi3kxzqajvx1qp1jz7az747y98-julia-1.8.3/share/julia/stdlib/v1.8/Test/src/Test.jl:1360 [inlined]
> [4] top-level scope
>   @ /gnu/store/ms5wncdgzv1p90xahj6pk6xk8lfx4c6a-julia-colortypes-0.11.1/share/julia/loadpath/ColorTypes/test/types.jl:170

I don't actually do julia programming so I wasn't able to try to propose a patch for this. I've attached the full build log.

[julia-colortypes-0.11.1.log](https://github.com/JuliaGraphics/ColorTypes.jl/files/10226614/julia-colortypes-0.11.1.log)

kimikage:
I have added unit tests for aarch64 using QEMU(#301 and #303).
As for v0.11.5 on nightly build, it passed the aarch64 test.
```
  Julia Version 1.12.0-DEV.406
  Commit 8f6418ea507 (2024-04-25 04:04 UTC)
  Build Info:
    Official https://julialang.org/ release
  Platform Info:
[...]

PR #303, opened 2024-04-25
Backport "Update workflows (#301)”

kimikage:
cf. https://github.com/JuliaGraphics/ColorTypes.jl/pull/301#issuecomment-2077022254


kimikage:
cf. #288

PR #305, opened 2024-04-26
Fix unstable `rand` tests

kimikage:
We do not have a way to obtain a stable multi-dimensional array of random numbers. That situation does not change with `StableRNGs`.
I think it is sufficient here just to make sure that the API works.

### DIFF
--- a/test/operations.jl
+++ b/test/operations.jl
@@ -190,6 +190,7 @@ end
     eltypes = (N0f8, N2f6, N0f16, N2f14, N0f32, N2f30, Q3f12,
                Float16, Float32, Float64)
     @testset "rand($C)" for C in (
+            Gray{Bool}, # issue #275
             (Gray{T} for T in eltypes)...,
             (RGB{T} for T in eltypes)...,
             AGray{Float32}, GrayA{Float64},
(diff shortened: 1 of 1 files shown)


kimikage:
From another perspective, ColorTypes.jl also does not need to guarantee the stability of random numbers (colors).

kimikage:
It seems that `macos-latest` has moved from `macos-13` to `macos-14`. 
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories
```
Julia Version 1.12.0-DEV.426
Commit 9f37ba86093 (2024-04-28 01:14 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: macOS (x86_64-apple-darwin22.4.0)
  CPU: 3 × Apple M1 (Virtual)
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, westmere)
Threads: 1 default, 0 interactive, 1 GC (on 3 virtual cores)
```
However, the binary I am appears to be for x86_64 macOS 13.

Anyway, we may need to reconsider the targets.

kimikage:
cf. https://github.com/JuliaMath/FixedPointNumbers.jl/pull/290

PR #307, opened 2024-05-06
Remove redundant pre-register StyledStrings compat

tecosaur:
Now that StyledStrings is registered, we can do away with this code (and also include StyledStrings support in a release, which would be nice).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -14,13 +14,13 @@ StyledStringsExt = "StyledStrings"
 
 [compat]
 FixedPointNumbers = "0.5, 0.6, 0.7, 0.8"
-# StyledStrings = "1" # FIXME: workaround for unregistered StyledStrings
+StyledStrings = "1"
 julia = "1"
 
 [extras]
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
-Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f" # FIXME: workaround for unregistered StyledStrings
+StyledStrings = "f489334b-da3d-4c2e-b8f0-e476e12c162b"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Documenter", "Pkg", "Test"]
+test = ["Documenter", "StyledStrings", "Test"]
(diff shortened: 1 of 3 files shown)


kimikage:
Note that there is also "test/Project.toml".

tecosaur:
Hmmm, looks like I might need to make a patch release to the compat version of `StyledStrings` to improve the behaviour when precompiled within another package.

tecosaur:
Yup, I'll do a 1.0.1 release shortly: see https://github.com/JuliaLang/StyledStrings.jl/commit/dc94cfaf159f5e5018c21be28242df31ecd846f7

kimikage:
I will merge PR #308 first to make it easier to see the CI results. (The problem on ARM is expected to be resolved with FixedPointNumbers v0.8.5.)
Also, I believe Dependabot will be creating a PR in a couple of days, but it is not a priority.

tecosaur:
Cool, I'd thought I'd have 1.0.1 out by now, but there seems to be a holdup to do with the casing of `juialang` in the repo. I'll update this PR once that's resolved.

kimikage:
This is off-topic and outside my scope of responsibility, but what is the review process and accountability of StyledStrings.jl?
Committing without PR deprives us of the opportunity to review.
Of course, GitHub allows people to comment on commits, but that goes unnoticed by most people.

tecosaur:
The way StyledStrings.jl is setup: Commits to Main go through PRs, always, even for small fixup commits that really don't need it. Commits to other branches (like `julia1-compat`) can just be pushed.

Currently, I'm hoping that other people will get involved with this, but it's pretty much a just-me show ATM. Given that, the review process is "I think it looks good".

kimikage:
https://github.com/JuliaGraphics/ColorTypes.jl/blob/e1954578012621c02f9607181474bc48121661d2/test/runtests.jl#L24-L29
Could we remove the `try` block?

tecosaur:
I don't see why not, or why it was added in the first place?

kimikage:
> why it was added in the first place?

Because, as you know, `julia1-compat` didn't work before.

kimikage (approved):
LGTM. 
(However, a separate decision on the release plan is needed.)

PR #313, opened 2024-10-17
Tag 0.12 and remove old compat bounds

ViralBShah:
Removing very old versions of FixedPointNumbers.jl in compat, and prepare to tag v0.12.0.

cc @dgleich


ViralBShah:
I'll plan to register this in the next week.

PR #319, opened 2025-02-27
explicitly qualify `Tuple` while adding method

nsajko:
Prevent name ambiguity, adapt to changes in nightly Julia, preventing a warning during precompilation.

See:
* https://github.com/JuliaLang/julia/issues/25744
* https://github.com/JuliaLang/julia/issues/57290
* https://github.com/JuliaLang/julia/pull/57311

PR #321, opened 2025-03-26
Qualify Tuple in method overload

serenity4:
This removes the [following warning](https://github.com/JuliaDebug/Cthulhu.jl/actions/runs/14084243275/job/39444048842?pr=630#step:7:150) on 1.12+:

```
  1 dependency had output during precompilation:
┌ ColorTypes
│  WARNING: Constructor for type "Tuple" was extended in `ColorTypes` without explicit qualification or import.
│    NOTE: Assumed "Tuple" refers to `Base.Tuple`. This behavior is deprecated and may differ in future versions.`
│    NOTE: This behavior may have differed in Julia versions prior to 1.12.
│    Hint: If you intended to create a new generic function of the same name, use `function Tuple end`.
│    Hint: To silence the warning, qualify `Tuple` as `Base.Tuple` in the method signature or explicitly `import Base: Tuple`.
└  
```

which, as indicated, should be avoided in favor of an explicit overload.

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -133,4 +133,4 @@ coloralpha(c::C) where {C<:TransparentColor} = coloralpha(base_color_type(C))(co
 coloralpha(c::C,a) where {C<:TransparentColor} = coloralpha(base_color_type(C))(color(c), a)
 
 # Tuple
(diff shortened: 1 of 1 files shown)


nsajko:
duplicate of #319

serenity4:
Thanks, should probably have checked other PRs first lol

PR #326, opened 2025-11-16
Add the CMY color model

andreuvall:
This PR adds the [CMY color model](https://en.wikipedia.org/wiki/CMYK_color_model#CMY) (the "essential" variant of CMYK).

This is one of my first PRs, and for sure the first in the family of Julia color packages. It goes along with https://github.com/JuliaGraphics/Colors.jl/pull/574 and https://github.com/JuliaGraphics/ColorVectorSpace.jl/pull/198). 

I was hoping to get some feedback before doing more work on this, to make sure I do it in the right direction. For example, I am aware of the following (and please let me know other things that I am missing):
- Although I am defining `CMY{T<:Fractional}` as a parametric type, I am not too sure how to handle this parameter information in the definition of related methods
- I probably should add tests
- I could add information on the CMY type in ColorTypes.jl/README


timholy:
Seems like you're on the right track to me. Since we already support `CMYK`, I think it's clear that this package can support both additive and subtractive models.

andreuvall:
Thank you for the feedback, @timholy!

Although I had taken a look, I didn't realize that the Julia color packages support CMYK. Now that I have searched again, I have found occurrences of CMYK in [`customtypes.jl`](https://github.com/JuliaGraphics/ColorTypes.jl/blob/0f4b4aba8adb7ab60371750ed0c78d8afcc64096/test/customtypes.jl#L86). I have also seen these interesting [thread](https://discourse.julialang.org/t/pre-rfc-ann-color-related-packages-yuvcolortypes-cmykcolortypes-colorprofiles-hdrcolortypes-hdrimages/113032) and [package](https://github.com/kimikage/CMYKColorTypes.jl).

Reading the linked thread, I see confirmation for my decision to implement CMY instead of CMYK. For me, CMY is enough for the printing application I am working on (because of its subtractive-ness), while being straightforward to convert from and to RGB. The conversion is seemingly not simple for CMYK.

When I find time, I'll keep on working on it. What I can think of is
- [ ] adding tests and
- [ ] updating `ColorTypes.jl/README` to reflect the addition of CMY.

I'll let you all know when I am done with that.

kimikage:
Support for 3-component `CMY` is a step in the right direction, but it has the potential to cause significant confusion.
It seems undesirable for `CMY` and `CMYK` to exist in different type hierarchies and exhibit different behaviors.

PR #328, opened 2026-01-28
Fix GPU kernel compilation by replacing @pure with @generated functions

habemus-papadum:
## Summary

ColorTypes uses several `@pure`-annotated functions to perform compile-time
type computations (walking the type hierarchy, extracting parameter bounds,
checking element type support). This worked on Julia 1.9 with GPU compilers
(CUDA.jl, Metal.jl) because `@pure` allowed the compiler to constant-fold
these calls, eliminating all dynamic dispatch from the generated IR.

Starting around Julia 1.10, `@pure` was deprecated in favor of
`@assume_effects`. More critically, the GPU compiler infrastructure
(GPUCompiler.jl) disables semi-concrete interpretation to avoid issues
with method table overlays (JuliaLang/julia#47349). This means the
compiler can no longer fold away the recursive type-walking in
`_eltypes_supported` and the subtype check (`<:`) in `issupported`, leaving
dynamic dispatch calls (`jl_apply_generic`, `jl_f_issubtype`) in the LLVM IR.
Metal.jl rejects these as `InvalidIRError`:

- `unsupported dynamic function invocation (call to _eltypes_supported)`
- `unsupported call to an unknown function (call to jl_f_issubtype)`

See: https://github.com/JuliaGPU/Metal.jl/issues/613

## What was tried

**Attempted fix:** replacing `@pure` with `@assume_effects :foldable` on all five
annotated functions. This did **NOT** resolve the issue — the compiler still
failed to fold the recursive type-hierarchy walk during GPU compilation.

**Working fix:** convert `eltypes_supported` and `issupported` to `@generated`
functions. Generated functions execute their body during compilation and
return a quoted expression, guaranteeing the result is a compile-time
constant with zero runtime dispatch. The recursive implementation is
moved to `_eltypes_supported_impl` helper functions that are called inside
the `@generated` body. The remaining three `@pure` sites (`to_top`, `basetype`,
`pureintersect`, `_parameter_upper_bound`) are replaced with
`@assume_effects :foldable` as a modernization, though these are not on
the critical GPU code path.

## Test plan

- [x] Metal.jl MWE (`@metal kernel(MtlArray{RGB{Float32}}(undef, 1))`) runs without `InvalidIRError`
- [x] Full ColorTypes test suite passes (6226 tests, 0 failures)
- [ ] CUDA.jl code path calls the same ColorTypes functions — should be compatible but was not tested directly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/ColorTypes.jl
+++ b/src/ColorTypes.jl
@@ -2,7 +2,17 @@ module ColorTypes
 
 using FixedPointNumbers
 import FixedPointNumbers: floattype
-using Base: @pure
+@static if VERSION >= v"1.8"
+    using Base: @assume_effects
+    macro pure_or_foldable(ex)
+        esc(:(@assume_effects :foldable $ex))
+    end
+else
+    using Base: @pure
+    macro pure_or_foldable(ex)
+        esc(:(@pure $ex))
+    end
+end
 
 const Fractional = Union{AbstractFloat, FixedPoint}
 
--- a/src/traits.jl
+++ b/src/traits.jl
@@ -120,7 +120,7 @@ color(c::TransparentColorN{5,C}) where {C} = C(comp1(c), comp2(c), comp3(c), com
 # recurse up the type hierarchy until you get to Colorant{T,N} for
 # specific T,N.
 to_top(::Type{Colorant{T,N}}) where {T,N} = Colorant{T,N}
-@pure to_top(::Type{C}) where {C<:Colorant} = to_top(supertype(C))
+@pure_or_foldable to_top(::Type{C}) where {C<:Colorant} = to_top(supertype(C))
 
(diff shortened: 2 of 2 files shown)


kimikage:
The `@pure` annotation here has existed as a necessary evil. Since it is a misuse, it should ideally be removed, and if possible, you should avoid using `@assume_effects` as well.
（cf. #208)

Although a fundamental reconsideration and survey are needed, merging this PR first is one option.

kimikage:
It's possible that the reason the Attempted fix didn't work was due to code coverage(cf. issue #339, PR #340).

Since `@generated` inherently requires purity, it potentially carries the same kind of issues as the misuse of `@pure`.

